PR #441, opened 2014-12-03
Move filters to the right side.

Typz:
This keeps all navigation links on the left, while the board filters are
moved to the right side.


fguillot:
I introduced a drop down menu now to be able to add more filters.

Issue #235, opened 2014-09-02
remember filter


proserver1:
is: apply a filter -> edit task -> go back to board == filter does not apply anymore
expected: I'd expect the filter to be still applied

if I filter, I usually do it like "show me all category X tasks". do this so long untill no task is open anymore or until I decide to change the filter...

fguillot:
I just merged a pull-request about that: https://github.com/fguillot/kanboard/pull/487

PR #487, opened 2015-01-05
Update board.js to always keep filter settings #235

gavlepeter:
Save filter settings to localStorage everytime filter_apply() is called.

Get and set filters from localStorage when filter_load_events() is called.

This improves the user experience when browsing tasks on a filtered board. No need to set the filters each time the board loads.

Perhaps a checkbox option to initialize this functionality but I think it should be default behavior.

### DIFF
--- a/assets/js/board.js
+++ b/assets/js/board.js
@@ -213,6 +213,14 @@ Kanboard.Board = (function() {
                 item.style.opacity = "0.2";
             }
         });
+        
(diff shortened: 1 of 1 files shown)


fguillot:
Awesome, I just modified your patch to prefix the keys with "filters_".

I just discover the double bitwise in javascript `~~(filterDueDate)` :)

Thanks.

PR #504, opened 2015-01-11
Implemented Project duplication with Tasks included within Controller\Project

mluepkes:
Included task-duplication withing project duplication as requested in fguillot#344 and needed in my own setting. Right now tasks are ALWAYS copied. Will open issue for future settings and options


mluepkes:
Closed and up for ReDo by implementing a ProjectDuplication model

Issue #92, opened 2014-05-16
Support linking of tasks together


oliviermaridat:
I would like to suggest a new feature: the ability to link too tasks together. If the link as a meaning (e.g. "related to", see below for more), it is possible to explain the relation between too tasks and organize better the work. This also simplify the browsing between tasks.

Examples of links:
* #1 Related to #7
* #1 Waits for #7 (and #7 prevents #1)
* #1 Subtask of #7 (and #7 parent task of #1)

Technically, a link contains: the source task id, the link meaning, the target task id. Each task should be able to have several links (#1 is a subtask of #8 but is also the parent task of #3 and is related to #17).
 If a link is created on #1, it appears on the task, for example as below:
![2014-05-16_555x368_scrot](https://cloud.githubusercontent.com/assets/953989/2998720/2ad733f4-dd0e-11e3-9ff9-31249b099585.png)
and the inverse link appears automatically on the linked task.

What do you?

oliviermaridat:
It should be fixed by now:
* A task can't be linked to itself anymore
* The "array to string conversion" error seems to be removed 
* I have added the duplication of links when a project or a task is duplicated. For a task duplication, the link label of the original project is used. Note that when editing the new duplicated task links, only the link labels of the current project are available (which is quite logical). Actually, this is similar to linking 2 tasks from 2 different projects: it works, but one project is always using the link labels of the other project. I don't know yet if this is a real issue.

oliviermaridat:
I am currently preparing the pull request, so:
* I have made some enhancement and clarification in the UI to show / edit links
* it is not possible anymore to have duplication of links
* I have added a French translation for the link feature
* [oliviermaridat/kanboard/master](https://github.com/oliviermaridat/kanboard) is now up to date with fguillot/kanboard/master (wow, loads of updates!).
* I didn't test this for Postgres...

It seems fine for me now. I will do a final check this week and then send a pull request.

BlueTeck:
I will test it again later ;)

BlueTeck:
I am not sure if I made is mistake or not ;) Got the following message: 

Fatal error: Class 'SimpleLogger\Logger' not found in /var/www/virtual/########.de/htdocs/todo_dev/app/ServiceProvider/LoggingProvider.php on line 15

BlueTeck:
SimpleLogger is not in my Vendor Folder...

oliviermaridat:
"composer.lock" problem is now solved.
[...]

PR #510, opened 2015-01-12
Support linking of tasks together

oliviermaridat:
As described in  #92, this pull request adds the ability to link too tasks together in order to be able to explain the relation between too tasks and therefore organize better the work. This also simplify the browsing between tasks.

* A list of link labels per project can be managed through the Project view
* A list of default link labels for new project can be managed through the Config view (Board settings)
* A link to an other task can be added in the Task show view
 * Linking two tasks of the same project is the common behaviour
 * Linking two tasks of different project is working, but only the link labels of the current project are available during the link creation
* The number of links to other tasks is displayed at the bottom of each task in the Board view
* Project links are duplicated when a project is duplicated
* Task links are duplicated when a task is duplicated
* This PR has been tested for SQLite and MySQL but not for Postgres
* This PR includes English and French translations only


oliviermaridat:
Sorry for the high number of commits, I don't know (yet) how to handle that more properly.

fguillot:
Awesome, give me the week to test and review that.

index.php
```diff
@@ -1,5 +1,7 @@
 <?php
-
+ini_set('display_errors',1);
+ini_set('display_startup_errors',1);
+error_reporting(-1);
```
fguillot: error_reporting should not be hardcoded

app/Template/tasklink/edit.php
```diff
@@ -0,0 +1,37 @@
...
+
+    <div class="form-actions">
+        <input type="submit" value="<?= t('Save') ?>" class="btn btn-blue"/>
+        <?= t('or') ?>
+        <?= $this->a(t('cancel'), 'task', 'show', array('task_id' => $task['id'], 'project_id' => $task['project_id'])) ?>
+    </div>
+</form>
+<?php else: ?>
+<div class="alert alert-info">
+    You need to <?= $this->a('add link labels', 'link', 'index', array('project_id' => $task['project_id'])) ?> to this project before to link this task to another one.
```
fguillot: This text cannot be translated.

app/Template/link/index.php
```diff
@@ -0,0 +1,49 @@
...
+		<i><?= t('#10 Follows #9') ?></i>
+	</div>
+    <?= $this->formCsrf() ?>
+    <?= $this->formHidden('project_id', $values) ?>
+
+    <?= $this->formLabel(t('Link Label'), 'name') ?>
+    <?= $this->formText('name', $values, $errors, array('required', 'autofocus', 'placeholder="Precedes"')) ?>
+
+    <?= $this->formLabel(t('Link Inverse Label'), 'name_inverse') ?>
+    <?= $this->formText('name_inverse', $values, $errors, array('required', 'placeholder="Follows"')) ?>
```
fguillot: placeholders are not translated

app/Model/TaskLink.php
```diff
@@ -0,0 +1,342 @@
...
+                    return false;
+                }
+            }
+        });
+    }
+
+    public function isUniqueConstraintFailed()
+    {
+        return $this->search($this->db->getLogMessages(), 'UNIQUE') || $this->search($this->db->getLogMessages(), 'Duplicata');
+    }
```
fguillot: You should not do that. That will work only in debug mode. Check if the link exists at the validation.

app/Controller/Tasklink.php
```diff
@@ -0,0 +1,181 @@
...
+                'task_id' => $task['id'],
+                'another_link' => $this->request->getIntegerParam('another_link', 0)
+            );
+        }
+
+        $this->response->html($this->taskLayout('tasklink/create', array(
+            'values' => $values,
+            'errors' => $errors,
+            'link_list' => $this->link->getList($task['project_id'], false),
+            'task_list' => $this->taskFinder->getList($task['project_id'], Task::STATUS_OPEN, $task['id']),
```
fguillot: I got a fatal error here, should be "TaskModel::STATUS_OPEN"

fguillot:
Why the links are specific per project, why not global to the application? That simplify a lot of stuff. 

Categories and swimlanes are specific per project because they can be different but you will always define the same link types "blocked by/blocks this task...", it's like task colors for me.

oliviermaridat:
Thank you very much for all your comments. I checked and checked my code, but apparently not enough. I will see that asap.

> Why the links are specific per project, why not global to the application?

I made them specific per project because I thought they may be different depending of business activities (online shops project will have different links than development project) like categories. So, not exactly like task colors in my opinion. 
But for sure this add more complexity. This is your call, what do you prefer?

fguillot:
Jira have those links:

link_jira

I think those "relations" should be there by default, that cover most needs.

I would prefer to have links global to the application, by example available from a new section in the settings page.

My concern is you have done a lot of works to make your pull-request, are you ok to change that? Maybe someone else have a different opinion? after all it's an open source software.

oliviermaridat:
Normally I have fixed all the issues you spotted. But the PR will fail, because I am using my proposed PR for [simpleValidator](https://github.com/fguillot/simpleValidator) (not to force your hand of course, task link validation can be done differently in kanboard if you prefer). 

This will give me the time to think about "global links" ;-) I am very open and I am ok to change whatever you want. I just want to do it the best possible way.

oliviermaridat:
I am working in a different branch ([oliviermaridat/kanboard/link](https://github.com/oliviermaridat/kanboard/tree/link)) to make link labels global to all project. Actually, this doesn't change many things in the code, but it will clearly simplify the management for the end-user. Anyway, it will still be possible to add a "link label" for a specific project. This offers simplicity and specificity :-) But if you don't like "specificity", feel free to remove it.
My only regrets is that I can't add that properly to the Controller/Config (too much pages to create), so I keep it in Controller/Link even it is accessible through the application Settings page.

I am also working on clarifying the "link" and "link label" data models. I was never completely satisfied with it, but I didn't find how to do that properly. I have a new idea, so I will try it.

I will send a new PR when this will be done.

BlueTeck:
I like it :+1: 

Note: The default links should be translated.

fguillot:
I close this pull-request, just open a new one when you will be ready.

Thanks.

Issue #544, opened 2015-01-23
formatting fails for URLs with numeric fragment identifiers


norcnorc:
If I write a URL like

http://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454

Into a kanboard task description, the resulting HTML is invalid, because kanboard is trying to format it as two kinds of links at once: A link to stackoverflow, and a link to task number 1732454

fguillot:
Fix merged.

PR #545, opened 2015-01-23
fix invalid HTML when linking to URLs with numeric fragment identifiers

norcnorc:
Fix for #544.

The code tried to convert task numbers to links by applying a regex substitution to the HTML output of Parsedown. This can sometimes break the HTML.

One example is links to URLs with a fragment identifier that starts with a number, such as http://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454

The fix is to add the additional processing to the markdown renderer itself, so it's only applied to appropriate parts of the text, and never to the generated HTML tags.

### DIFF
--- a/app/Core/Helper.php
+++ b/app/Core/Helper.php
@@ -5,6 +5,34 @@ namespace Core;
 use Pimple\Container;
 use Parsedown;
 
+class _ParsedownKanboard extends Parsedown
+{
+    public function __construct($link, $helper) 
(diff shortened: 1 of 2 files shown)


fguillot:
Thanks, I merged your patch.

I moved the inherited Parsedown class to a separate file and adapted the code to the last stable version of Parsedown.

PR #546, opened 2015-01-23
Removed inactive projects, their tasks and subtasks from dashboard

mluepkes:
Dashboard does no longer display disabled projects and their respective tasks and subtasks.


fguillot:
Thanks it's merged, I also added the method `getActiveMemberProjects()` to be able to filter events from the dashboard.

PR #548, opened 2015-01-24
Added column description in Task Boards.

ashbike:
Each column in a board may have description which is free form and textual in nature. This field can be used for many purposes like:

* A general description of this column.
* Entry criteria: When a task qualifies to move to this column.
* Checklist for exiting this column.

This feature adds a column 'column.description' of type text. One can add/edit descriptions to columns from [project]-> Edit Board link.

In the board view, one icon appears next to the column name if a description was added to it. A pop-over on that icon shows the description. Markdown is supported in description however, Write/Preview options are not shown.

### Screenshot
image

### DIFF
--- a/app/Controller/Base.php
+++ b/app/Controller/Base.php
@@ -312,4 +312,23 @@ abstract class Base
 
         return $project;
     }
+    
+    /**
+     * Common method to get a column
+     *
+     * @access protected
+     * @param  integer      $column_id    Default column id
+     * @return array
+     */
+    protected function getColumn()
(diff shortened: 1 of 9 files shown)


ashbike:
I get Travis::The command "php phpunit.phar -c tests/units.sqlite.xml" exited with 2.

I guess something got broken in JSON API, I need to check that. Any help on this is appreciated. Will re-initiate the pull request after fixing this.

ashbike:
Will resend this pull request.

PR #549, opened 2015-01-24
Added column description in Task Boards

ashbike:
Each column in a board may have description which is free form and textual in nature. This field can be used for many purposes like:

*  A general description of that column.
* Entry criteria: When a task qualifies to move to this column.
* Checklist for exiting this column.

This feature adds a column 'column.description' of type text. One can add/edit descriptions to columns from [project]-> Edit Board link.

In the board view, one icon appears next to the column name if a description was added to it. A pop-over on that icon shows the description. Markdown is supported in description however, Write/Preview options are not shown.

*The earlier commit was creating Travis > The command "php phpunit.phar -c tests/units.sqlite.xml" exited with 2.*. Updated the test case and locally running phpunit does not give me any error. Hope it will get through.


ashbike:
I see it fails again. NOt sure what's causing this. Any help will be appreciated.

fguillot:
I merged and changed some stuff.

ashbike:
Hi @fguillot, I see that markdown is gone after changes. But I guess that should be there since you already have the markdown processor. How about this?

1. In the Board Edit page, the table shows all boards, name, description etc. Instead of description, we will show the description icon there. Hovering on the icon one can see the description.
2. The edit column will have the markdown capability with preview.
3. Back to the board, one can hover on the column name and get the formatted description. However, I would like to have a small indicator there showing description is available. Otherwise user will not know about it. And all the columns might not have description, user will only know after pointing to it.

Let me know your thoughts.

Issue #543, opened 2015-01-23
predictable task URLs


norcnorc:
In kanboard 1.0.8, the URL "/kanboard/?controller=task&action=show&task_id=42" would open task number 42. That was not as nice as #118 would like, but not too bad.

In version 1.0.11, we need "/kanboard/?controller=task&action=show&task_id=42&project_id=666". If the project_id value is missing or not correct, then the link won't work, because [this check](/fguillot/kanboard/blob/4e874fad90832cf325ed8b718ae55c58dd92b1db/app/Controller/Base.php#L289) fails.

That makes it needlessly difficult to generate links to kanboard from external tools.

Can we go back to a URL scheme where knowing just the task number is enough to be able to generate a link?

fguillot:
Actually, the project_id is used to check project permissions...

mluepkes:
To be exact: the project_id url parameter is needed in ```Controller/Base::beforeAction()```. ```Controller/Task::show()``` just checks if the given task is contained within the given project.
Could be solved with adding an OR check in the beforeAction to either pull the project_id directly if given or get it via the task. Will see if I can fit in a fix tonight.

mluepkes:
Note: I just redid the check mechanism. The urls are unchanged, the project_id can be omitted though.

fguillot:
It's modified in master, if you omit the project_id an extra sql query is made to fetch the project_id from the task.

PR #552, opened 2015-01-26
Before filter

mluepkes:
Changed authentication check to either pull project_id from url if given or from task if only task_id is given as per #543


fguillot:
Your patch doesn't work, when you display any page without a "task_id" param that return a 404, that works only for tasks related pages. You have to check if the "task_id" param is there to get the project_id from the task.

I made this change directly in master.

Thanks.

PR #559, opened 2015-01-29
Allow public access to colors stylesheet.

crwilliams:
The CSS stylesheet at ?controller=app&action=colors is currently not available if the user is not logged in.

As a result of this, any public boards viewed in read-only mode to a user who is not (or has not been) logged in do not display the task cards in color.

This is a change to the ACL to allow access.


fguillot:
Thanks, it's merged.

PR #568, opened 2015-02-01
Added percent completed to the main view when a task has subtasks. Addre...

levlaz:
...sses https://github.com/fguillot/kanboard/issues/174

This works more or less. I think we should discuss if we want to keep the X/Y completed tasks or just replace it completed with the percentage. In addition, is it necessary to say " X% Completed" or is "X% sufficient?"

### DIFF
--- a/app/Template/board/task.php
+++ b/app/Template/board/task.php
@@ -98,7 +98,7 @@
     <div class="task-board-icons">
 
         <?php if (! empty($task['nb_subtasks'])): ?>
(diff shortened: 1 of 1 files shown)


fguillot:
Thanks, it's merged. I just keep the percentage.

levlaz:
Awesome! Thanks so much for accepting it! :) 

I have a quick question, I tried to look at doing this on the main task page as well (underneath the task-bar header) in templates/subtask/show.php but it looks like the array that is passed on this page is not the same array that is passed on that page. I tried to dig around but was not able to figure it out. Is it even worth doing on that page, if so, where is this variable defined. If we include nb_subtasks and nb_completed_subtasks then adding it to this page would be very simple as well. 

Best, 
Lev

fguillot:
Yes, it's not the same array, but something similar can be done with a template helper function. You have to count manually the number of subtasks completed and return the percentage.

levlaz:
Thank you for the clarification, I was starting to do just that, I will work on it a bit more and submit a PR. :)

Issue #575, opened 2015-02-02
Inconsistencies in duplicate functions / transaction handling


mluepkes:
While working on expanding project duplication to contain swimlanes and tasks, I've run into differences between how duplicate functions handle transactions. 

For some models (e.g. tasks) the duplicate functions also duplicate their sub-models. This means they're starting transactions of themselves. Others only handle their own duplication and call on db->save directly.

My question now is, do I:
* refactor all models to only handle their own duplication and superset the submodel and transaction handling.
* work in some sort of improved transaction handling?

The latter would also include changes to PicoDB to include an inTransaction function and returns for startTransaction and closeTransaction.

I think a combination of both would probably be of most benefit for the project in the long run.

fguillot:
For the Swimlane, just add a method `duplicate()` like other stuff (categories...)

The task duplication should be optional during project duplication, the user should be able to choose or not to duplicate tasks with a checkbox by example. So you can do it in two pass without modifying too much the current code:

1. Duplicate the project with all properties (one transaction: `ProjectDuplication::duplicate()`)
2. If the project is duplicated successfully, duplicate all task (`TaskDuplication::duplicateToProject()`)

mluepkes:
I didn't get the idea of doing it in 2 passes, the rest was already implemented. I guess that's workable, although ultimatively I'd prefer the ability of doing it in one pass and do a rollback if task duplication fails. I'll implement it this way for now. Just need to include the checkboxes for selecting which parts are to be duplicated and I'll post a pull request.

mluepkes:
Everything implemented, writing the unit test

PR #580, opened 2015-02-03
Implemented Changes to Project Duplication to include Swimlanes and Tasks.

mluepkes:
ProjectDuplication::duplicate accepts additional param of type array now. Array includes which optional parts to duplicate. Optional parts are: 'swimlane', 'category', 'task', 'action'.

### DIFF
--- a/app/Controller/Project.php
+++ b/app/Controller/Project.php
@@ -297,7 +297,7 @@ class Project extends Base
     /**
      * Duplicate a project
      *
-     * @author Antonio Rabelo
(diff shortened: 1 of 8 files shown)


mluepkes:
#575 

Also: *** Please generate new language files, had to use one new string. *** Could you include some sort of short introduction to the usage of the shell scripts to the manual?

fguillot:
It's merged, thanks.

For the translations, I updated the doc: https://github.com/fguillot/kanboard/blob/master/docs/translations.markdown

mluepkes:
Thank you!

Issue #596, opened 2015-02-09
[Bug] subtask text in board


BlueTeck:
The  popup in the board shows status instead of title:
![kb5](https://cloud.githubusercontent.com/assets/1961634/6105495/00527a4e-b058-11e4-9bbf-1f98ab59d62e.PNG)
![kb6](https://cloud.githubusercontent.com/assets/1961634/6105496/0052cdf0-b058-11e4-8b1e-8f238832be77.PNG)

fguillot:
Fixed. Thanks.

Issue #595, opened 2015-02-09
[Bug] add new Column to board


BlueTeck:
Add a new colum:
![kb1](https://cloud.githubusercontent.com/assets/1961634/6103589/4b22dfee-b046-11e4-9444-d40c188b3dfa.PNG)
Will add the description as task limit:
![kb2](https://cloud.githubusercontent.com/assets/1961634/6103597/5ac40eb4-b046-11e4-95f0-90f4698aac7b.PNG)
![kb3](https://cloud.githubusercontent.com/assets/1961634/6103598/5ac42fd4-b046-11e4-8ba7-7607473f2787.PNG)
Editing the column does not show tasklimit or description:
![kb4](https://cloud.githubusercontent.com/assets/1961634/6103615/7f742708-b046-11e4-96d3-79c17c995d83.PNG)

fguillot:
Fixed. Thanks.

PR #598, opened 2015-02-09
Miscellaneous enhancements fixes for column description

ashbike:
- Markdown supported in column description.
- Columns show a small right aligned 'info' icon if description available.
- Project > Edit Board and Project Summary pages now show description as tooltip.
- Project > Edit Board shows textarea with markdown preview. Edit column also shows markdown preview.
- Fixed bug while adding a new column inserts description in column limit. #595
- Column limit is always checked and zeroed via a intval (...) function. 
- Fix for subtask text in board #596

Note: php phpunit.phar -c tests/functionals.sqlite.xml reports failures (Tests: 44, Assertions: 58, Failures: 43.). But that happens with master. So I assume I can ignore it for now.

Demo can be found at: http://goo.gl/Y5PrgO

image

### DIFF
--- a/app/Controller/Board.php
+++ b/app/Controller/Board.php
@@ -283,7 +283,7 @@ class Board extends Base
 
         if ($valid) {
 
-            if ($this->board->addColumn($project['id'], $data['title'],$data['description'])) {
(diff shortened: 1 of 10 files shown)


fguillot:
Merged c92aae6

PR #601, opened 2015-02-09
added getProjectActivity + getOverdueTasks to API

BlueTeck:
- bind getProjectActivity and getOverdueTasks to API
- added start and end parameter to projectActivity->getProjects


fguillot:
Merged.

Issue #608, opened 2015-02-10
Task age and time spent in one column


ashbike:
In our internal implementation, two things we needed to track: 1. How long a task is open 2. How much time a task spent in a column (both in days).

I did a prototype which can be seen at: http://goo.gl/Y5PrgO. Screenshot below. There are tooltips on those little boxes next to the "Assigned to XXX" text on task board. The time spent in a column resets if the task is moved to another column.

Similar feature is available in other software, http://goo.gl/kptUdK . Also, I was referring to the slides at http://goo.gl/j6fTk9. They also propose something like this.
Kanboard code is so nicely organized, I had to add just 10-12 lines to have this feature. Let me know if you think this is useful. Then I can clean up the code and request for a pull. 

![image](https://cloud.githubusercontent.com/assets/7455931/6134734/75ff21c0-b189-11e4-9a52-773b842c18d5.png)

sparkles645:
That's cool!

fguillot:
It's definitively useful, you can send a pull-request :)

BlueTeck:
Idea: display hours, if the number of days is lower then 0.

ashbike:
Was thinking about that. How about <1h, NNh if less than <24, and finally days?

BlueTeck:
:+1:

fguillot:
I merged your pull-request.

If you want to display the hours, just send another pull-request :)

PR #612, opened 2015-02-11
Task age and time spent in one column.

ashbike:
This adds two display fields on board against each task. #608

- Task movement to other column resets the 2nd field.
- Task movement across swimlanes in same column does not change it.

This adds one field in 'tasks' table which keeps the last moved timestamp. Another SQL populates this field from 'project_activities' table for older tasks. For newly created tasks, this field is set to creation date/time.

*Note: I do not have a PGSQL database handy. So do not know same SQL will run against PGSQL or not. MySQL/SQLite uses same SQL so I hope it works in PGSQL*


fguillot:
Thanks it's merged.

PR #615, opened 2015-02-13
Task age and days in column show hours if < 1 day

ashbike:
Sending this pull request though I'm not very happy with this code change because of these reasons.

- Put the function in app/functions.php. Not sure if this is the right place. 
- Now it does not do any dynamic formatting. It's always <1h, NNh and NNd. 

It's functional, but I think the code can be improved.

### DIFF
--- a/app/Template/board/task.php
+++ b/app/Template/board/task.php
@@ -69,8 +69,8 @@
         ) ?>
     </span>
     
-    <span title="<?= t('Task age in days')?>" class="task-days-age"><?= t('%dd', floor(time()/86400) - floor($task['date_creation']/86400)) ?></span>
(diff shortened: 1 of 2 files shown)


BlueTeck:
:+1:

mkresin:
> Now it does not do any dynamic formatting.

You might want to have a look at the following code from frédéric's other project miniflux https://github.com/fguillot/miniflux/blob/master/lib/helpers.php#L137

ashbike:
:+1: That's what I was looking for. New to PHP. Any advice which file this function will go into? Then I will make changes and send another pull request.

fguillot:
I merged and modified your pull-request (move the function to the right place and the text can be translated).

Note: Template helper functions go to `app\Core\Helper.php`.

PR #617, opened 2015-02-13
Branding Kanboard - Adds a hook to add custom header. You can put logo etc.

ashbike:
We wanted to had some sort of branding (basically logo) on our Kanboard. I am deploying Kanboard for someone else and they too, were asking if their logo can be added somehow and some other style changes. Then thought of generalizing it. May be this is useful to others as well.

** By default it does not do anything and there is no visual change ** 

 You need to add a file "site/header.php". There is a "header.php.default" file under "site" directory, that can a starting point. Rename "header.php.default" to "header.php". Refresh Kanboard to see the effect.
By default Kanboard favicon shows up here. The file "site/css/style.css" also gets included in the pages. You can define your own classes here to use in "header.php". [And perhaps you can override default Kanboard styles :)]

image


BlueTeck:
:+1:

crash5:
target="_blank" is not better than target="new"?

ashbike:
Don't know why some work in progress got into it... closing and creating a new pull request.

ashbike:
@crash5 Correct. I don't know from where this "new" came in :). Fixed and created a separate pull request.

PR #618, opened 2015-02-14
Branding Kanboard - Adds a hook to add custom header. You can put logo etc

ashbike:
We wanted to had some sort of branding (basically logo) on our Kanboard. I am deploying Kanboard for someone else and they too, were asking if their logo can be added somehow and some other style changes. Then thought of generalizing it. May be this is useful to others as well.

** By default it does not do anything and there is no visual change **

You need to add a file "site/header.php". There is a "header.php.default" file under "site" directory, that can a starting point. Rename "header.php.default" to "header.php". Refresh Kanboard to see the effect.
By default Kanboard favicon shows up here. The file "site/css/style.css" also gets included in the pages. You can define your own classes here to use in "header.php". [And perhaps you can override default Kanboard styles :)]

Thanks @crash5 for pointing to the target = "new" thing. 

Note: Resubmitting the pull request since I messed up with Git. Some work in progress files got committed.


creador30:
Simply adding support for custom php should be sufficient. You can link custom css from within this php. 
For example, I use custom php to link different css files for normal view and public access.

fguillot:
I merged the pull-request #666 to have custom css. In theory, you can do the something similar by overriding the default css.

ashbike:
Just tried out the new release. Works good with custom styles. :+1:

PR #629, opened 2015-02-16
This adds a quick comment creation link in the task menu on board.

ashbike:
In our environment it's one of the most frequently used action. It may be useful to others as well.

*Just a thought, do we need 'Change assignee' and 'Change Category' in the menu? One can open them by clicking of the 'assigned' name or 'category' box. Then the menu is less crowded*
image


BlueTeck:
But if there is no category, you can't click it ;)

ashbike:
That's right. Category link should be there.

PR #644, opened 2015-02-20
display column in link details

BlueTeck:
Add column name to link list, see below:

kb1


fguillot:
I modified your pull-request because there is an issue when you link a task from another project with different column names than the current task project.

BlueTeck:
sry i didn't know that it is possible to link between projects.

PR #647, opened 2015-02-22
A small patch which enables scroll view toggling.

ashbike:
If there are many columns, the board adds a horizontal scrollbar. However, This looses the full board visibility. Now there is a menu item on the Board, 'Action > Toggle view' which switches. This can also be toggled by 'c' keypress.

**NOTE: I think the Javascript code can be reduced.  jQuery(document).ajaxComplete... those can be bypassed. I see the taskview toggling (stack_toggle ()...) keeps state. Couldn't find out how it keeps state after ajax refresh with my very limited knowledge of JS.**


fguillot:
What about using css class instead of using inline css? by example "board-compact-view" and "board-scroll-view" or something else.

ashbike:
:smile: I am asking the same question to myself. Why I didn't do that?

ashbike:
Created another pull request [ #668 ]  with changes suggested.

PR #656, opened 2015-02-24
Display footer icons left of text.

Typz:
This is consistent with the due date indicator, and needed to avoid confusion between the due date
and the following number of comments/tasks/...

IMHO this also feels slightly more natural :-)


ashbike:
:+1:

PR #657, opened 2015-02-24
Add project description field

Typz:
Description is shown in a popup when hovering the 'information' icon, like column description,
which is shown in project list on dashboard and in board's title.

This is useful to document the categories and colors used in the project.


ashbike:
:+1:

BlueTeck:
:+1:

fguillot:
I'm not sure if including the html content inside the title attribute was a good idea. There was a bug with Chrome/Firefox when you add a link. I fixed that by using a single quote for the attribute.

PR #659, opened 2015-02-24
Use same tooltips for column description and task footer.

Typz:


fguillot:
Do you have an idea why I cannot do a mouseover on tooltips for column description but I can do it with task footer tooltips?

Typz:
The mouseover part needs a bit of JavaScript event handling, which is done only for task-board-tooltip class, in board.js. The column use the  column-task tooltip, which is setup in base.js.

Adding the task-board-tooltip class changes this behavior on the board only (and defaulting to the title attribute in the content function).

Some refactoring would be needed to support this everywhere, but I am not sure if that would feel natural for a tooltip: that's why I did not go in that direction.

On the other hand this is probably needed if the description includes links...

PR #660, opened 2015-02-24
Avoid flicker when displaying the page.

Typz:
Use &lt;span&gt;/&lt;span&gt; instead of &lt;ul&gt;/&lt;il&gt; for task menu anchor, to avoid flickering items when the menus are created.


fguillot:
I never notice any "re-layout", how I can reproduce the issue?

Typz:
With chrome at least, on a large board (500 tasks) there is some flickering / relayout: there is a placeholder for the task menu's bullet, which eventually is removed when dropit has finished.

PR #664, opened 2015-02-24
Add new actions to move issue when the assignee is changed.

Typz:
This will *not* fire the task_moved events, to avoid triggering other auto-assign actions.
This does not seem ideal, but I see no easy way to implement this properly (e.g. checking the event
stack).

### DIFF
--- a/app/Action/TaskMoveColumnAssigned.php
+++ b/app/Action/TaskMoveColumnAssigned.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace Action;
+
+use Model\Task;
+
+/**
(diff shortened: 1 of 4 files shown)


fguillot:
What is the use case for that, do you have an example?

Typz:
I have New and InProgress columns.
I already have actions which automatically sets the assignee to the person who moves the task to InProgress, and clears it when the task is moved back to New.

This makes it more complex for manager to have somebody work on a new task (e.g. in meeting): with these actions, simply setting the assignee in column New will move the task to InProgress (and conversely)

PR #668, opened 2015-02-25
A small patch which enables scroll view toggling. - Replaces #647 

ashbike:
If there are many columns, the board adds a horizontal scrollbar. However, This looses the full board visibility. Now there is a menu item on the Board, Action > Toggle view which switches. This can also be toggled by 'c' keypress.

Replaces #647


ashbike:
This has got some problem which messes up dragging tasks. Will fix and test more.

PR #675, opened 2015-02-28
Toggle view

ashbike:
Added a menu item 'Toggle View' under Board -> Action and short key 'c' to toggle board view. View modes are compact (tries to stuff all columns in the window) and wide (scrollbar appears)

Replaces #647 #668

### DIFF
--- a/app/Template/board/filters.php
+++ b/app/Template/board/filters.php
@@ -14,6 +14,10 @@
                             </span>
                         </li>
                         <li>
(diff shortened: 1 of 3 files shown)


Typz:
This patch makes the page flicker when loading, since the column are not properly size at first, then the correct table "layout" is set by Javascript.

Also, the 'public' view is not properly sized at all, with all column having different width.

Issue #646, opened 2015-02-21
Tagging tasks - a prototype


ashbike:
We started missing tagging/labelling feature in Kanboard :smile:. Tried to come up with an implementation.

There is a basic version running on our demo setup http://goo.gl/Y5PrgO. You can add/edit tags, and they show up on board. No fancy tag input ui as of now. Just put your tags separated by comma. Tags can only contain a-z, 0-9, . <dot> <whitespace>, - <hyphen>,  _< underscore>. Always comma separated and lower case. It will remove all special characters, empty tags etc. Screenshot below.

### Caveat 1

* We already have colors and categories in Kanboard. They are system defined, and one can be selected at a time. 
* Tag/labels are user defined can have multiple values and somewhat free form.

As we see it, they complement each other.

*Note: In our implementation we color coded tags. It's not dynamic, rather controlled by css. You can specify common tags with their color codes in css. However, with tasks already colored, the board gets unpleasantly colorful. :smiley: .  

### Caveat 2

About the implementation, we thought of doing it in couple of ways. First choice was to do it in very formal way, normalized data model etc. One TAGS table, one TAG-TASK mapping table. But implementation got too complex very soon. Two more tables to join and lack of an standard aggregate functions to simplify things, extra queries while inserting, deleting etc. etc. It would have offered good flexibility, but then we thought if we really need that kind of complexities. Finally aborted. [ https://github.com/ashbike/kanboard/tree/tagging-orig ]

Other option was very simple, keep tags comma separated and put in the task table itself. WIth some string manipulation and stereotype queries, most of the functionalities can be achieved. Tag based grouping by can be done at PHP level. Not at database level. There will be some limitation but 90% can be achieved. So we took this route. [ https://github.com/ashbike/kanboard/tree/tagging ]

![image](https://cloud.githubusercontent.com/assets/7455931/6315073/821b92ea-ba1c-11e4-8c9f-8b87ee799590.png)

oparoz:
That looks really good and would be very useful when using hooks. but I'm a little bit worried about robustness. How about storing something like a serialized object?

About colours, I think it could work quite well if the main colours were replaced by shades of one colour per example, but overall, it's not too shocking. Pastel colours could maybe be used for the main tasks.

BlueTeck:
Looks nice, but i think categories could be removed then, and it should be possible to filter/search by tags.

ashbike:
[...]

PR #679, opened 2015-02-28
Prototype pf tagging

ashbike:
This refers to #646


fguillot:
The template 'task/tags.php' seems to be missing in the pull-request.

fguillot:
Tags will be probably implemented in the future but differently (not with the "csv" way).

Thanks for making this prototype.

Issue #307, opened 2014-10-04
Preview image/pdf on the card


fguillot:
Feature request from the Twitter account: https://twitter.com/gruenler/status/518108481036255232

lawre:
+1 on the request for an image thumbnail on the kanban card.  This feature is the *only* reason why I haven't ditched Trello yet.

BlueTeck:
What do you think about this?

![kb_img](https://cloud.githubusercontent.com/assets/1961634/6430242/a233ba2c-c001-11e4-955b-af5bc8e7da7a.PNG)

ashbike:
:+1:

oparoz:
Looks good. Which image formats does it support?


   FROM: Timo
SENT: Sunday, 1 March 2015 10:57
TO: kanboard@noreply.github.com
REPLY TO: fguillot/kanboard
CC: Olivier Paroz
SUBJECT: Re: [kanboard] Preview image/pdf on the card (#307)

> What do you think about this?
>
> [1]
>
> —
> Reply to this email directly or view it on GitHub[2].



Liens:
------
[1]  
https://cloud.githubusercontent.com/assets/1961634/6430242/a233ba2c-c001-11e4-955b-af5bc8e7da7a.PNG
[2] https://github.com/fguillot/kanboard/issues/307#issuecomment-76590221

BlueTeck:
at the moment kanboard supports image types, that are supported from php getimagesize https://github.com/fguillot/kanboard/blob/master/app/Model/File.php#L122 http://php.net/manual/de/function.getimagesize.php

This php command supports:
[IMAGETYPE_GIF] 
[IMAGETYPE_JPEG] 
[IMAGETYPE_PNG] 
[IMAGETYPE_SWF] 
[IMAGETYPE_PSD] 
[IMAGETYPE_BMP]
[IMAGETYPE_TIFF_II] 
[IMAGETYPE_TIFF_MM] 
[IMAGETYPE_JPC] 
[IMAGETYPE_JP2] 
[IMAGETYPE_JPX]
[IMAGETYPE_JB2] 
[IMAGETYPE_SWC] 
[IMAGETYPE_IFF] 
[IMAGETYPE_WBMP] 
[IMAGETYPE_JPEG2000] 
[IMAGETYPE_XBM] 
[IMAGETYPE_ICO]

BlueTeck:
I have changed the layout, i think it looks much clearer, but you could see the whole width of the picture, but not the full height.
[...]

PR #684, opened 2015-03-02
preview images in taskview

BlueTeck:
enhance attachment view in task details, part of #307 

Screenshot: 
https://github.com/fguillot/kanboard/issues/307#issuecomment-76602088


BlueTeck:
I will make a new pull request, with better code that is using thumbnails.

PR #688, opened 2015-03-03
Fix flickering and column sizes.

Typz:
* Avoid flickering in default view mode, by implementing the default/horizontal-scroll view mode
properly. This flickering was introduced with #675 pull request.
* This also fixes the layout of the public board.
* Use fixed-layout on the table to improve layout and rendering speed.
* Slightly reduce the width of the swimlane header column, and make it fixed-width.
* Slightly simplify the Javascript/jQuery code for activating compact mode.

There is still some flickering in compact mode, as the mode is implemented directly by Javascript
code after the page is loaded.


fguillot:
The page header seems to be broken when you enable the horizontal scrolling the header don't use the full page width.

Typz:
Oups, I did not see that...

This can be fixed by simply putting back the 'overflow-x: auto' line on the #board-container element. However, this causes the scrollbar to appear 180px below the board, thus looking quite strange:
    * When the board has few task, the scrollbar appears somewhere in the [vertical] center of the browser window
    * When the board has many task and vertical scrollng is needed, the scrollbar appears only after scrolling all the way down

To avoid all this I also tried to use 'fixed' positioning of the header (which also ensures the headers are not moved when Chosen replaces the controls; this works great as far as horizontal scrolling is concerned, but also retains its fixed position when scrolling vertically: it does not move, and gets 'below' the board.

I updated my change with this 'fixed' layout, but I am not sure how it can be properly fixed.

Typz:
Closing this, please check #706 instead: it has only the table layout part, not any changes to the scrolling. I will see if i can fix the scrolbar issues (e.g. horizontal scrollbar appears in the 'middle' of the page if board is empty, and must scroll to the bottom of the page to get the scrollbar when there are many tasks).

PR #692, opened 2015-03-04
Add category description.

Typz:
It is shown in tooltip over the category indicator on the board, and helps documenting what each
category means.


fguillot:
I think we need to reduce the number of sql queries to display the board. The category name and description are in the same table, that can be fetched in a single query. After that there is probably a way to improve the variables passing to the template.

Typz:
Updated to get name & description in a single query, and use the 'categories' variable to pass everything to the template.

fguillot:
Merged and modified.

PR #697, opened 2015-03-06
Changed maxlength of subtask titles to 100.

okraits:
A maxlength of 50 for subtask titles is too short in my experience. A maxlength of 100 should be sufficient in most cases - if it is not sufficient you should think about splitting your subtask further more :-)


BlueTeck:
Did you check if the database field is long enough?

okraits:
I didn't find any restriction regarding the database column. The validation in app/Model/Subtask.php checks for a maximum length of 100.

Issue #669, opened 2015-02-25
Simplify the task link creation


oliviermaridat:
The new task link functionality is available since a few days now (thanks fguillot for the merge, the work on simplification and all the enhancements you add!). I am using it a lot, and I find it useful but quite boring to use when several task links have to be created. (creation of link labels is ok)
I would like to propose the following enhancements to facilitate the task link usage.

* Add an item "Add a link" in the dropdown menu of a task in the board view
* Add a quick task link creation as it is possible for subtask
* Add the possibility to update a task link
 * The problem here is that it may add complexity. It was available in my original PR, but fguillot prefers to simplify the whole functionality (and I can understand that!). But it is so boring to remove AND create a new task link for just changing the link label of an existing one. I assume it is possible to find an easy way to do that.
* Add the task number in the autocompletion search during a task link creation

What do you think about that? Maybe someone have already some feedbacks to give about this new feature.

oliviermaridat:
In my first PR to answer this ticket, I remarked that the part about "grouping" task link by labels has been discard. If possible, I would like to discuss this point.

At the moment, to display task links, we have something similar to:
![link1](https://cloud.githubusercontent.com/assets/953989/6666996/5e1934ae-cbe9-11e4-967c-fef5b0c4678e.png)

But I would rather prefer to have something like that:
![link2](https://cloud.githubusercontent.com/assets/953989/6666997/5f445a84-cbe9-11e4-809c-fbcabea73b9d.png)

I find it cleaner and more readable. It is also easier to copy/paste a quick overview of a milestone (or any other link) to somewhere else (e.g. a Word/Openoffice report).

I know that the UI is all about feelings. So I will completely agree if you disagree with this ^^ But, just to know if this is a disagreement on the feature, or on the way to implement it.

fguillot:
The grouping is a good idea, it's more readable.

fguillot:
To be tested in master

PR #704, opened 2015-03-10
Simplify the task link creation

oliviermaridat:
In order to simplify the task link creation, I propose the following updates (partial implementation of #669):
* Add an item "Add a link" in the dropdown menu of a task in the board view
* Add a quick task link creation as it is possible for subtask
* Order task links by labels and task column


fguillot:
Thanks, it's merged.

PR #706, opened 2015-03-11
Fix flickering and column sizes.

Typz:
* Avoid flickering in default view mode, by properly setting the default (horizontal-scroll) layout
directly in CSS. This flickering was introduced with #675 pull request. This also fixes the layout
of the public board. Javascript is now used only to activate the 'compact' mode, thus still causing
a slight flickering, but in that mode only.
* Use fixed-layout on the table to improve layout and rendering speed.
* Slightly reduce the width of the swimlane header column, and make it fixed-width.
* Slightly simplify the Javascript/jQuery code for activating compact mode.

This is replaces pull request #688, with only the table layout fix, since I have no proper solution
for the scrolling problem.


fguillot:
I still having the same issue as before due to the fixed layout of the table:

hscroll_tablefixed

I don't see the flickering with my own boards, how many tasks do you have on your board?

Typz:
This is fixed now, the header does not move anymore.

I have a board with more than 400 tasks (luckily, most of them in 'Done' or 'Backlog' columns :-) ), and i cannot miss the resizing of columns with Chrome on my Linux machine.

The other problem with this "large" board is that I need to scroll down to the bottom of the page to see the horizontal scrollbar. This is the problem I tried to fix as well in #688, causing the header regression. This patch addresses only the board layout/flicker issue, to avoid the header regression.

fguillot:
Thanks it's merged.

I see your point about the horizontal scrollbar. I didn't notice that because I use a macbook and I can scroll horizontally with the touchpad without using the scrollbars.

PR #721, opened 2015-03-17
Add an action which logs the time a task has spent in a configured column

enko:
This is useful, to track the time a task has spent in your work column.

I just duplicated the TaskLogMoveAnotherColumn task and adapted it to my needs. I'm not so deep into the hole code base, so I dont know if I could have fetched the project activities somewhat smarter or if thats ok.

### DIFF
--- a/app/Action/TaskLogTimeSpent.php
+++ b/app/Action/TaskLogTimeSpent.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace Action;
+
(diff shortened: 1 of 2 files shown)


fguillot:
There is a new feature introduced in the current master, it's called "task transitions", that record all task column moves with the time spent in the origin column and the person that does the action.

fguillot:
Closed because there is task transitions now.

Issue #693, opened 2015-03-05
Activity log within the task detail


greg35:
Hi all,

It could be useful to have the full list of activities summed up in the task detail.
For the moment, when we opened a task we have :
[Task detail]
[Date & Time tracking]
[Description]
[Subtask]
[Comments]

Why not add a part [Activity] with the activity log (eg : John move the card to "In progress", Max close the Task....)

Or include these logs with the comments.

Thanks for you feedback

BlueTeck:
good idea

@greg35 There is already an automatic action, that adds a comment when the task is moved to another column

greg35:
OK I will try this feature thanks a lot for the information.

fguillot:
The activity stream is now available in the task view.

PR #746, opened 2015-03-29
add activity stream to task #693

BlueTeck:
add activity stream to task #693

activity


fguillot:
Merged and modified to avoid some code duplication and fix php 5.3 compatibility (don't use [ and ] for arrays).

PR #763, opened 2015-04-03
Ability to assign a task color based on an event.

manishlad:
Initially supported event: task moved to a column


fguillot:
I renamed the class to `TaskAssignColorColumn` to be more specific.

manishlad:
Yes, I was debating whether to make it more specific or not :-)
Anyway, that's great - Thanks for the merge.

Issue #717, opened 2015-03-16
[FR] File attachment add size, date uploaded, user uploaded


BlueTeck:
Enhance the information about uploaded file:
- add file size
- add the user who uploaded the file
- add date of upload

If @fguillot agree, i will implement that and make a PR.

Regards,
Timo

fguillot:
Yes, no problem.

PR #779, opened 2015-04-08
fixed preview images size on hover

lyoshaprikhodko:
Size changing on hovering is now fixed. (hope that was not a feature)

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -1091,7 +1091,7 @@ span.task-board-date-overdue {
 }
 .task-show-images li .img_container {
     width: 250px;
-    height: 100px;
+    height: 190px;
     overflow: hidden;
 }
 .task-show-images li {


BlueTeck:
It was a feature :-)

semka:
It is very annoying feature :)

Chiliec:
I agree, it's very uncomfortable when you want to open image in modal window and button run away from cursor.

Konshin:
agree ^

BlueTeck:
Okay, if you want, we could remove it, but this pull request ist not the right way, i could do it by my own. i will change it with the #717 enhancement next week.

semka:
Awesome, thanks a lot. Our testing team goes mad every time they need to click on preview :)

BlueTeck:
I have created a PR for that #782

fguillot:
I merged #782

Issue #775, opened 2015-04-07
Add Time tracking in linked task view


oliviermaridat:
It could be useful to have a column "Time tracking" (as it is done for subtasks) in the "Links" table of a task view. This simplify the visualization of time spent on a task. 

Typical use case:
* One task is a milestone and therefore have a link "is a milestone of" with several other tasks
* The manager of this milestone will add subtasks ("Management", "Validation", ...) to estimate and track the time spent around this task.
* The manager can see quickly the time spent on the subtasks of this milestone, and on the tasks that targets this milestone.

oliviermaridat:
FYI: I am working on a PR for this.

fguillot:
See master.

PR #785, opened 2015-04-09
Simplify task link creation and usage

oliviermaridat:
In order to simplify the task link creation and usage, I propose the following update (to continue the implementation of #669):

* Add update task link feature
* Add time tracking and assignee columns in the Links table of the task view (see #775)
* Group the task links by label in the Links table of the task view (see #669)

Now, the Links table is like that:

links-table

I had to update also the task id autocomplete process to make it work for updates. Also, the previous implementation prevent the creation / update of a task link without Javascript. Now, without Javascript, a simple task id field is displayed. With Javascript, this task id field is hidden, the task title autocompletion field appears, and the "Save" button is disabled.


oliviermaridat:
Hi @fguillot, if you intend to merge this PR but if you need something more (or less) to do it, please, don't hesitate to ask me.

fguillot:
Sorry about the delay, I will try to review/merge your pull-request this week-end.

fguillot:
I don't understand something, what is the purpose of `opposite_task_link_id`, why do you need to modify the schema? you already have `opposite_task_id`? That doesn't seems to be used anywhere, only at task link creation...

oliviermaridat:
task_id and opposite_task_id are used in a task link to specify the two linked tasks. E.g. task 1 blocks task 2 / task 2 is blocked by task 1.

Whereas opposite_task_link_id is used to specify a relation between two task link tuples. Like "opposite_id" for link labels. Actually, this relation can be retrieved with the task_link.id, task_link.task_id and the task_link.opposite_task_id and links.id / links.opposite_id. But this involves a quite complex SQL request. Especially because link labels can be bidirectional (like "related to"). I tried that in my first PR but it was very complicated, and I suppose this is why you discarded it.
That is why I propose to add opposite_task_link_id, which is filled during the creation of the two task link tuples, but which is useful to update the opposite task link tuple. I thought it was easier to implement and to understand.

To be honest, I tried many data models for that, and it was at least the first that didn't involve complex SQL requests. It can for sure be done differently, and maybe better.

fguillot:
I merged manually your pull-request and changed some stuff:

- The label grouping looks like your initial ticket #669
- I removed the schema modification and use the unique index to find the opposite task link
- The task links in the board tooltips is the same as before, I will see later to show the same table as in the task view...

oliviermaridat:
Thanks fguillot! I tested it, and it seems to work just fine.
I really like the idea of having the same look and feel for links and subtasks in the task view. In the current implement they are completely different and I fear it is therefore quite confusing for the user. Do you think we can keep a column order similar to the subtasks one? Label, task, Column, Assignee, Time tracking, Actions

fguillot:
I prefer to keep the table like that for now especially when you have a small screen.

oliviermaridat:
Actually, if the subtask table fits small screens, the link table should fit also. If it get too large because of the "label" column, this column can be moved to a row (which has one bonus advantage: a label row can even summarize the time tracking of its tasks, which can be really useful for the blocks / targets milestone labels).
For readability, and because the concepts of subtask and link are really similar, I would prefer to display similar tables for subtasks and links in the task view.

Anyway, it is your project and your decision. I argue for the sake of the controversy, but please, do as your prefer :-)

PR #788, opened 2015-04-10
Small visual changes

EmersonMellado:
Hi,

It is my first collaboration on gitHub. I am excited about Kanboard, because I think it is a very good system. My intention is to help you on Issues, if it is ok.

This was a very small change, just to see how it all works.

Thank you for your time.

Emerson


fguillot:
Unfortunately, I cannot merge your pull-request, your fork is too old, your are 55 commits behind the current master.

The Brazilian translation is almost up to date, check here:  https://github.com/fguillot/kanboard/blob/master/app/Locale/pt_BR/translations.php

If you want to add a shadow and rounded corner around the card you are better to use the custom stylesheet, that doesn't require any code modification. By the way, don't modify directly the `app.css` this file is auto-generated, see http://kanboard.net/documentation/assets.

Thanks.

PR #798, opened 2015-04-12
Update translations.php

gavlepeter:
Also add translation for Swedish currency: 
'SEK - Swedish Krona' => 'SEK - Svensk Krona',


fguillot:
Thanks, it's merged.

PR #809, opened 2015-04-16
Fixed a bug where deletion of columns could break the 'Move up/down' column functionality

nedelenbos:
Hi,

We recently encountered a bug that we couldn't move a column upwards in one of our boards. After some researching it seems that when a column is removed, the sort order gets messed up. Instead of getting a list of position 1, 2, 3, the order can be (ex;) 1, 2, 4 after deleting. 

When moving the item on the fourth position up, a PHP notice UNDEFINED INDEX occurs, because the position to move to is not taken anymore (it's removed, since the column was removed).

This fix ensures that the position is changed with a valid row. It loops through the positions until a valid position is found.

The safeguards for calling the moveUp/moveDown columns have also been changed appropriately, as well as the template where the 'Move up' / 'Move down' actions are shown.

### DIFF
--- a/app/Model/Board.php
+++ b/app/Model/Board.php
@@ -158,7 +158,7 @@ class Board extends Base
         $columns = $this->db->hashtable(self::TABLE)->eq('project_id', $project_id)->asc('position')->getAll('id', 'position');
         $positions = array_flip($columns);
 
-        if (isset($columns[$column_id]) && $columns[$column_id] < count($columns)) {
+        if (isset($columns[$column_id]) && $columns[$column_id] < max($columns)) {
             // Increase the position, for as long as we don't find another column
             do {
                 $position = ++$columns[$column_id];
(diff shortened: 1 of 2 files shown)


nedelenbos:
@fguillot 

Can you explain why my changes made the following test fail (only @ PHP 5.5)?

```
1) ProjectActivityTest::testCreation
Failed asserting that '1429178129' matches expected 1429178130.
/home/travis/build/fguillot/kanboard/tests/units/ProjectActivityTest.php:44
```

The assertion that goes along with this one is:
```PHP
$this->assertEquals(time(), $events[0]['date_creation']);
```

This error does not strike me as a failure due to my code, but rather due to unfortunate circumstances..

fguillot:
Thanks for reporting this bug.

I have done some cleanup/refactoring in the original code because the same issue was happening in subtasks. I also added a couple of unit tests for that.

Thanks again.

PR #818, opened 2015-04-20
No category quick

skelliam:
Hi Frédéric,

My patch is really simple -- the proposal is to always show "No Category" box on each item if there isn't a category.  This saves the user one click + wait time.  Since the category dropdown is a main feature of the board UI, it seems this should always be shown instead of hidden.

Old way: 
1. Click card, wait for new screen
2. Find and click "Edit Task"
3. Select new category
4. Save

New way:
1. Click "No Category"
2. Select new category
3. Save

With regards,

Bill


fguillot:
I understand your point but not displaying that save some real estate on the card when category are not relevant.

Issue #340, opened 2014-10-21
Recurring Tasks


aaronrobertson:
Could you add a feature to add new tasks to a project on a specific schedule?

majewsky:
This would interest me as well. I would like to use the board for recurring household tasks, such as:
* vacuum living room every week
* clean bath every three weeks

shimza:
Me too. Recurring tasks and location would be ideal

Cheers,

Shimon Friedman 
Mob: 0403067613

fguillot:
@majewsky did you already forgot to vacuum the living room? :)

hamzasheikh:
Another vote for this feature. I'm OK with creating n number of additional tasks when the task is first created. I can then edit their due dates to manage their order.

David-Norris:
Proposed concept: #847

fguillot:
I merged/improved the pull-request of @David-Norris in master.

The idea behind is described here: https://github.com/fguillot/kanboard/blob/master/docs/recurring-tasks.markdown

We will see later if that need some modifications or improvements.

PR #847, opened 2015-05-06
Recurring tasks concept (#340)

David-Norris:
In this concept, essentially, when triggered, a task marked as recurring is duplicated to the first column of the same board in the same swim lane and the due date (if there is one specified) is adjusted accordingly.

There are 3 triggers that currently create a new recurring task; moving a task from the first column, moving a task to the last column, or closing the task (trigger is configurable per task).

Due dates, if set on the current task, can be recalculated by a given factor of days, months or years. The base date for the calculation of the new due date can be either the existing due date, or the trigger date (configurable per task).

Each task tracks the task id of the parent task that created it, and the child task that it creates.

Once a task has created a child task, the recurring settings cannot be changed in the UI (can be updated in the API).


David-Norris:
@fguillot. Sorry, I meant to do the Vagrantfile change as a separate pull request. Hope this is OK.

fguillot:
Thanks for the pull-request, It's now merged in master! I improved some part of the code and added some unit tests...

David-Norris:
Thanks @fguillot for accepting/merging it in (and adding the unit tests and other housekeeping tasks). Sorry for some of the troubles encountered. For an enhancement, I specifically didn't do it, because I personally feel it goes against Kanban methodology, but I did have the idea if people need the ability to create multiple tasks upfront, it would be possible to add a new trigger status (Manual), and manually call duplicateRecurringTask in a loop to create many tasks in advance. With proper checks and balances, the loop can be for a either a set iteration (create 5 duplicate tasks), or until a specific due date is reached. Another enhancement would be instead of triggering from the first or last column or closing of the task, would be to add another trigger type and process it when moved to a column that has an automatic action setup. There were lots of ideas, however I wanted to get a base concept in first to see if/how it was accepted. Great work on Kanboard. Love it.

Issue #935, opened 2015-06-19
Error when deleting a subtask


thedragon013:
Hello,

when deleting a subtask in version 1.0.14, I get the following message:

Warning: sprintf() [function.sprintf]: Too few arguments in /xxxx/app/Core/Translator.php on line 52

The subtask is deleted, though. I use german as language.

Regards,
dragon013

fguillot:
It's now fixed in the development version with the patch of @crwilliams.

Thanks to report this issue.

PR #936, opened 2015-06-19
Fix translations with incorrect placeholders and add unit test.

crwilliams:
Includes fix for issue #935.


fguillot:
Thanks, so many bugs fixed in the translations!

PR #959, opened 2015-06-29
added the feature to open 'create another task' inside a popup

wiebkem:
Hello,

I added the feature to open 'create another task' inside a popup.


fguillot:
The same feature have been implemented in master but differently.

PR #962, opened 2015-06-29
added the feature to flag / unflag a task

wiebkem:
Hello,

I added the feature to flag / unflag a task.


fguillot:
What is the purpose of that? use case?

wiebkem:
As a user I want to flag / unflag a task to indicate that the task is important, but not completely clear at the moment.
If a task is flagged then the other users know that the task description is not completely done and clear to start working at it.

NinjaPewPew:
1.0.25 has added a priority attribute to tasks. That should be enough to signal someone that a task is important. Adding a Flag to task might just confuse people.

PR #1000, opened 2015-07-08
Enable nonAdmins to change private projekts to normal

deliciousbob:
- isAdmin function must be disabled to enable a manager to convert private projects to normal projects.
- Edit.php code is updated with isManager to enable Managers to update private projects to normal projects by editing in the project edit formular.

### DIFF
--- a/app/Controller/Project.php
+++ b/app/Controller/Project.php
@@ -140,7 +140,7 @@ class Project extends Base
         $project = $this->getProject();
         $values = $this->request->getValues();
 
-        if ($project['is_private'] == 1 && $this->userSession->isAdmin() && ! isset($values['is_private'])) {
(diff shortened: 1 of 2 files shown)


fguillot:
I keep your pull-request on hold because some changes about groups will be probably made in the future. It's not yet completely defined, but a new group at the application level called "super users" or "managers" might be introduced. Basically, the same thing as admins except that cannot change the application settings.

deliciousbob:
Hi Fred, no problem, if the new group is coming, my work around would be obsolete, so I'm looking froward to it. Kind regards Robert.

fguillot:
There is now a new group call "Project Administrator" in the dev version just for that.

See https://github.com/fguillot/kanboard/blob/master/docs/user-management.markdown

PR #1002, opened 2015-07-08
Sorted Colors and SelectList with BG Colors

deliciousbob:
I sorted the colors beginning by yellow to gray.
Added a css style -> option.color- and edited the form helper to output  class="color-..."

### DIFF
--- a/app/Helper/Form.php
+++ b/app/Helper/Form.php
@@ -53,7 +53,7 @@ class Form extends \Core\Base
 
         foreach ($options as $id => $value) {
 
-            $html .= '<option value="'.$this->helper->e($id).'"';
+            $html .= '<option class="color-'.$this->helper->e($id).'" value="'.$this->helper->e($id).'"';
 
             if (isset($values->$name) && $id == $values->$name) $html .= ' selected="selected"';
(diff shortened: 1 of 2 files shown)


fguillot:
That doesn't seems to work for me, at least on Google Chrome and Firefox. If I understand correctly your pull-request, the background color of the dropdown should be the selected color.

The css style should be applied on the select element when the value is changed, not the option?

deliciousbob:
Hi Fred, I tested it on my local version 1.0.16, where it worked fine.

I changed a thing in the get css function and just tried with the nightly build and it's now working fine.
I also changed the order of the colors, it should look like this in the edit and new form.
[color-change-kanboard](https://cloud.githubusercontent.com/assets/13135271/8670544/4c0741f0-2a1e-11e5-9e89-7828bbefa55c.jpg)

Best regards Robert

fguillot:
Ok that seems to work only on Windows platform.

The form helper have been modified that means all dropdown of the application will have a useless option element with a class "color-ID" that doesn't exists.

fguillot:
I changed the color selection in the development branch to use something that works on all platform: 

color-picker

Issue #971, opened 2015-07-02
Milestone and roadmap management


oliviermaridat:
I have a proposal for a better management of milestones through task links. I already have a proof of concept in my laptop, but I would like some feedbacks before to put more effort in it.
Let me explain my objectives and my proposal.

I would like to:
- have proper milestone and roadmap features
- simplify the task link view

I would also like to enhance the time tracking system. Actually, I know there is a lot of work about this at the moment, so I won't treat this now.
### Milestone and Roadmap

My idea is to treat "is a milestone of" links differently from the others. This will provide a better "Milestone" management, and clarify the Task Link view that was already too complex but still not enough complex to cover all use cases. Therefore, in the task view, instead of having:
- Links: list of link labels and their related tasks + time tracking per task
- Subtasks: list of subtasks + time tracking per subtask
- Time tracking

We could have:
- Milestone: list of tasks that targets this milestone + time tracking per task + total time tracking
  - (optionaly) A click on a task id displays the "dropdown menu", like in the board view
  - (optionaly) Each task is displayed as a collapsed colored card, like in the board view
  - (optionaly but difficult) Each task card can be drag and dropped to change the order
- Subtasks: list of subtasks + time tracking per subtask
  - (optionaly) + total time tracking
  - (optionaly) + list of tasks that has a link "is a child of" (they can be seen as a different kind of subtask)
- Time tracking (refactoring may be required, but this is another discussion ;-D)
- Links: a more simple list of link labels and their related tasks (without "is a milestone of")
  - Maybe some columns may be removed, like "Time tracking"...

This can be rendered like this:

![milestone](https://cloud.githubusercontent.com/assets/953989/8475686/1f6ded3c-20ba-11e5-96ad-e6022b2d7e91.png)
![links](https://cloud.githubusercontent.com/assets/953989/8475786/d9050140-20ba-11e5-9993-6076f7bb3023.png)

With this solution, the view of non-Milestone task is simplified, and the view of Milestone task provides better information.

After that, a new "Roadmap View" can be added to Kanboard. This view will display the list of all Milestones (i.e. Task linked as "is a milestone of") of a board and their related tasks, as I just described for the task view.

So what do you think of all this? Good idea / bad idea? Suggestions, questions... ?

fguillot:
Instead of using task links, why not build something just for managing milestones?

I look at several software and I like the simplicity of Github issue. You can create a milestone with a name, description and a due date. Any issue can be part of a milestone.

[...]

PR #1009, opened 2015-07-10
Milestone management

oliviermaridat:
This PR implements some of the idea about milestones listed in #971.

In the task view page, instead of having:

* Links: list of link labels and their related tasks + time tracking per task
* Subtasks: list of subtasks + time tracking per subtask
* Time tracking

we now have:

* Milestone: list of tasks that targets this milestone + time tracking per task + total time tracking
 * Each task is displayed as a collapsed colored card, like in the board view
 * A click on a task id displays the "dropdown menu", like in the board view (see second commit)
* Subtask: no change
* Time tracking: no change
* Links: a more simple list of link labels and their related tasks (without "is a milestone of")

The result is:

milestone_v2

I separate this in 2 commits:

* "Add special section for milestone in task view" only adds the milestone views
 * The "progress bar" use the color of the current task
 * The algorithm to compute the progress bar is: SUM_all tasks( min(time spent, time estimated) ) / SUM_all taks(time estimated)
* "Enable dropdown menu for other page than board" adds the task dropdown menu to the milestone view.
 * To implement that, I had to update all popover views to manage properly the redirection. Because the current behaviour is: ajax => redirect to board, otherwise redirect to task view. Now we can choose were to redirect after an update through a popover page. So, normally, we can now add the task dropdown menu in the listing view, calendar view (and the future roadmap view?) quite easily.


fguillot:
Can you make a separate pull-request for the task milestone view and the dropdown menu? that will be easier to review. Thanks.

oliviermaridat:
Ok, I will create a PR for dropdown menu only. I have another idea for the milestone view, which may fit better with "your" vision of task link. I will need some time to work on it.

PR #1039, opened 2015-07-19
Czech translation

rousarp:
incomplete, only important


fguillot:
Thanks for your contribution, Kanboard is now translated into 19 languages!

PR #1117, opened 2015-08-18
Norwegian translation

danielraknes:


fguillot:
Thanks for your contribution, Kanboard is now translated into 21 languages!

I get some issues with the file encoding since everything have to be utf-8 encoded, after tweaking the file everything seems ok.

PR #1198, opened 2015-09-06
Added initial version of Dutch (NL) translation

woutsanders:
Initial translation for Dutch users. I saw some more translation strings, so will take a look at those to make it 100% up-to-date (if possible). There are now still some translation strings missing (tested on latest stable version: 1.0.18


fguillot:
I think you have forked the wrong git repository. The version of Kanboard you have in your fork is too old. I cannot merge your pull-request because there are conflicts detected.

By the way, someone else already start to translate Kanboard in Dutch. You should update/improve the existing Dutch translations. (https://github.com/fguillot/kanboard/blob/master/app/Locale/nl_NL/translations.php)

woutsanders:
Hmm... Perhaps I overlooked something then, because I really thought that I had forked the current repository. I also couldn't find a dutch translation. I will have a look at that translation and will see whether it could be improved :)

Thanks for the quick reply though. I'm sorry if I caused any trouble.

patrickvanelk:
I tried to add some strings to the Dutch translation, first time ever :-P so I hope everything went ok.

PR #1220, opened 2015-09-12
fix margin of full-calendar header

julesverhaeren:
fix unnecessary margin below calendar header row, traced it's introduction back to https://github.com/fguillot/kanboard/commit/a04ecbde778decfdea7200806a6b1144861ae05f . It seems this was for the Last Logins page, but removing margin-bottom:20px does not seem to damage the layout of this page.


fguillot:
Your patch was causing some issues with the rest of the application, so I disabled the margin only for calendars. https://github.com/fguillot/kanboard/commit/b16e7d187216b2928611c2312627bcebcf35a2ab#diff-828a8364a616a0c62da41f5a71b8a9e3R10

PR #1221, opened 2015-09-12
do not show ugly scrollbar unless it is necessary

julesverhaeren:
overflow-x:scroll 
causes the public page to always show a horizontal scrollbar, even when this is not necessary
overflow-x:auto should fix this


fguillot:
Do not send the minified file, I cannot even see your modification, push the source css file so I can merge your pull-request (same for #1220). Thanks.

julesverhaeren:
My bad, hopefully I've fixed both pull requests now

PR #1241, opened 2015-09-17
[Docker] Add checkout cmd line to install last stable version

arcanneero:
With this, you sould add tag in docker hub for each version.


fguillot:
Now, there is 2 tags for docker images:

- stable: for the latest stable release
- latest: for the development version

Both images can be downloaded from Docker Hub, see https://github.com/fguillot/kanboard/blob/master/doc/docker.markdown#use-the-stable-version

PR #1347, opened 2015-10-09
add feature "move top" and "move bottom" into board columns settings

erwang:
Add the feature "Move top" and "Move bottom" into board column settings


fguillot:
The problem with your pull-request is your are calling continuously the method `moveDown` until that return false. That will generate a lot of SQL queries. There is probably a more efficient way to do that.

erwang:
i wanted to use the method moveDown to not duplicate code. I thought that the methods moveTop and moveBottom won't be often used and there won't be more SQL queries than the number of columns. If you think it's very negative, I'll make the changes.

fguillot:
In the development version, column positions can be changed by drag and drop now, there is no moveUp/Down anymore.

Thanks for your contribution.

Issue #1340, opened 2015-10-08
[Imp] Custom Filter option to select append/new filter


BlueTeck:
Idea to improve my custom filter function:

Add an option if a custom filter should reset the current filter, or if it appends the custom filter.

Reference: https://github.com/fguillot/kanboard/issues/1181

If I have some time, I will add this feature, maybe on the weekend. or someone else is faster ;)

lehrhardt:
that would be awesome...

thob:
:+1:

PR #1350, opened 2015-10-10
add append/replace option to custom filters

BlueTeck:
Issue #1340


app/Template/custom_filter/add.php
```diff
@@ -13,8 +13,10 @@
     <?= $this->form->text('filter', $values, $errors, array('required', 'maxlength="100"')) ?>
 
     <?php if ($this->user->isProjectManagementAllowed($project['id'])): ?>
-        <?= $this->form->checkbox('is_shared', t('Share with all project members'), 1) ?>
+        <?= $this->form->checkbox('is_shared', t('Share with all project members'), 1) ?>
```
fguillot: Remove the trailing spaces

BlueTeck: done

PR #1361, opened 2015-10-12
Fixes the javascript issue on swimlane board 

alirz23:
Since you already have handlers registered on the page you don't need to remove them and then add them again when popover opens up

### DIFF
--- a/assets/js/src/Popover.js
+++ b/assets/js/src/Popover.js
@@ -15,7 +15,6 @@ Popover.prototype.open = function(link) {
     $.get(link, function(content) {
         $("body").append('<div id="popover-container"><div id="popover-content">' + content + '</div></div>');
         self.router.dispatch();
-        self.app.refresh();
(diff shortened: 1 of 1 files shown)


assets/js/src/Popover.js
```diff
@@ -15,7 +15,7 @@ Popover.prototype.open = function(link) {
     $.get(link, function(content) {
         $("body").append('<div id="popover-container"><div id="popover-content">' + content + '</div></div>');
         self.router.dispatch();
-        self.app.refresh();
+        // self.app.refresh();
```
fguillot: Can you remove the line completely?

fguillot:
You don't need to change the permissions of the .htaccess for this fix.

alirz23:
permissions were for my machine I guess I checked in those files by mistake

alirz23:
I update the files

PR #1385, opened 2015-10-18
Fix readonly description tooltip

TheIndifferent:
In "Public view" the app.js was not included, so description tooltip was not shown. This patch makes the tooltip visible on the readonly view.

First attempt was to just reuse the same app.js everywhere, but I was afraid that I will include some unwanted functions that way, plus the page size was considerably bigger. So extracted everything into separate script, got 241 kB instead of 519, and everything appears to be working.

I have exactly 0 experience with PHP and very little experience with javascript, so please review carefully and comment if there's something to be fixed.


fguillot:
Javascript is voluntary not loaded in the public board view. Why do you need to have description tooltip working if the public view? the description still visible by clicking on the task.

TheIndifferent:
Because when there are lots of tasks and lots of swimlanes you have to switch to collapsed view, otherwise board becomea hard to navigate. In the collapsed view description tooltip is very helpful.

fguillot:
Need to be rebased.

Issue #1370, opened 2015-10-15
[Improvement] Choose wich colors will be visible and usable


oxydron:
It's a very nice feature to have square for colors to colorize tasks, but, in most teams I think that 3 or 4 colors will be used the most to represent categories or importance levels.

If we could set which colors will be visible, would be nice.

Best regards,
Ben Hur

sparkles645:
@oxydron 
Doesn't this request kind of contractict your original request - ref #1071 ?

* We use "Categories" and "Automatic actions" to decide the colors of the task - **would that solve your use case?**

BenWShade:
And choose the exact colours in settings. We have colour blind people on the team and they can't tell the difference between 80% of the current default colours so is useless to them. If they could edit the default colours to make them 'more different' then all the team could use the colours fully.

Great job so far. 
Thanks

oxydron:
@BenWShade  :+1: even better indeed.

@manuelflara :+1: Good work.

Great job guys!

manuelflara:
I tried to address both of your concerns when writing the code for my pull request. 

Hopefully it fulfills your needs :)

PR #1405, opened 2015-10-23
Customize Colors in Settings Page

manuelflara:
This is a pull request to add functionality requested in issue #1370.

Basically, a new Settings > Custom Colors page has been created. Each color allows for its background, border and name (what you see when hovering over the color in Create Task, also for filtering by color name) to be customized. Also, each color has a checkbox labeled "Selectable?" that allows colors to not even be shown at all when creating a task. So if you only want to define and use a handful of colors, you don't need to see all sixteen colors.

Upgrade instructions / compatibility details: it should all work without a hitch after updating your code base (see details below). 

1. I tried to be as backwards compatible and to change as few things as possible to accomplish this task. The most important decision was not changing how tasks.color_id is defined in the schema. Basically, I don't know how other people or tools might be relying on that column being a text based color name, and that string being one of the hard coded colors in Kanboard. So I chose to work around that instead of implementing a better data structure.

2. I opted for populating the `colors` table at runtime based on Model\Color::default_colors instead of having those inserts hardcoded in the schema migration files. This way it will adapt to local modifications (custom colors) any project may have when upgrading.

3. I left calls to getDefaultColors instead of the new getColors in a couple of places where I thought it made sense. For example, color projects on the Grantt formatter I think are better suited with the default pastel palette. Let me know if you'd rather I change this so user defined colors are used everywhere.

4. I chose to use a VARCHAR as the primary key for the new colors table, and choose the defaultColors' key as the ID, for consistency with tasks.color_id. I understand this isn't very proper, and it may not even make sense semantically later on (a 'yellow' color_id with a background of #ff0000 after user customizes it), but again, I tried to add this feature changing as few code as possible and trying not to break anything in existing installations.

5. In order to also retain the exact same order of colors when creating a task (for UX-backwards compatibility reasons), and since this is currently defined by the color's order in the defaultColors array, I added a `colors`.`position` column that I order the results by.

6. Having said that, if you'd rather I rework this so it's properly done even if it that means changing how tasks.color_id is defined, let me know.

7. I left Api\App::getColorList as is since I believe it's used to get a list of "available colors" from the API by tools that integrate with Kanboard. So I guess the current behavior is appropriate: get a list of custom colors that are usable, and not return those marked as not selectable by the user in the new Custom Colors page.

8. I added a few new strings to the translations files, including Spanish translations since they're the ones I can vouch for. I added them to the rest of the files as comments since that seems to be the MO for strings that don't have yet a translation.

9. I work with MySQL so I'm not sure if the migrations I created for sqlite and Postgres are totally correct. Also, the full database dumps on app/Schema/Sql/ seem like they are autogenerated (actual dumps) so I didn't add the definition of the new table `colors` to them. Should I?

10. In terms of managing data of these custom colors, the order in which they appear to the user in the Create Task view can be altered by changing the values of `colors`.`position` manually in the database. And to force a "reset to default colors", one can just truncate the colors table entirely.

11. Functional tests: I believe UserApi::testGetColorList already tests enough, since customizing colors is not available through the API so colors can't be changed through it.

12. Unit tests: I added a new test case for Model\Color as there wasn't any and that Model class underwent a lot of changes. Writing unit tests isn't my forte so I tried to mimicking similar tests on the code base.


app/Model/Color.php
```diff
@@ -100,6 +107,15 @@ class Color extends Base
     );
 
     /**
+     * Colors used in the app, since defaults can be overriden in the database.
+     * Each color also contains bool 'is_usable' array key.
+     *
+     * @access private
+     * @var array
+     */
+    private $colors = null;
```
fguillot: You can default to array() instead of null so php knows the type at variable creation.

app/Model/Color.php
```diff
@@ -108,9 +124,12 @@ class Color extends Base
      */
     public function find($color)
     {
+        if (!$this->colors) {
+            $this->loadColors();
```
fguillot: Instead of calling multiple times the method `loadColors()` you can add constructor that override the parent one and then load the colours from the database.

Example:

```php
public function __construct(Container $container)
{
    parent::__construct($container);
    $this->loadFromDatabase();
}
```

app/Model/Color.php
```diff
@@ -177,6 +204,30 @@ public function getDefaultColors()
...
+        }
+        if (!$only_usable_colors) {
+            return $this->colors;
+        }
+        $usable_colors = array();
+        foreach ($this->colors as $color_id => $values) {
+            if ($values['is_usable']) {
+                $usable_colors[$color_id] = $values;
```
fguillot: You can probably simplify this loop by using the function `array_filter()`.

app/Model/Color.php
```diff
@@ -210,16 +261,102 @@ public function getBackgroundColor($color_id)
...
+            $this->loadColors();
+        }
         $buffer = '';
 
-        foreach ($this->default_colors as $color => $values) {
+        foreach ($this->colors as $color => $values) {
             $buffer .= 'div.color-'.$color.' {';
             $buffer .= 'background-color: '.$values['background'].';';
-            $buffer .= 'border-color: '.$values['border'];
+            $buffer .= 'border-color: '.$values['border'].' !important';
```
fguillot: What is the purpose of `!important`? (yes I'm a dick in css)

app/Schema/Postgres.php
```diff
@@ -5,7 +5,21 @@
...
 use Kanboard\Core\Security;
 
-const VERSION = 73;
+const VERSION = 74;
+
+function version_74(PDO $pdo)
+{
+    $pdo->exec("
+        CREATE TABLE colors (
+            color_id VARCHAR(50) NOT NULL,
```
fguillot: The primary key is missing for Postgres.

app/Schema/Postgres.php
```diff
@@ -5,7 +5,21 @@
...
+            name VARCHAR(30) NOT NULL,
+            background VARCHAR(25) NOT NULL,
+            border VARCHAR(25) NOT NULL,
+            is_usable SMALLINT NOT NULL DEFAULT '1',
```
fguillot: Use the type `BOOLEAN` instead.

app/Schema/Mysql.php
```diff
@@ -5,7 +5,22 @@
...
 use Kanboard\Core\Security;
 
-const VERSION = 93;
+const VERSION = 94;
+
+function version_94(PDO $pdo)
+{
+    $pdo->exec("
+        CREATE TABLE colors (
+            color_id VARCHAR(50) NOT NULL COMMENT 'Backwards compatible format with Kanboard color IDs.',
```
fguillot: I'm not sure if the comment if the schema itself are really necessary. The column names are easily understandable.

app/Schema/Mysql.php
```diff
@@ -5,7 +5,22 @@
...
+            name VARCHAR(30) NOT NULL COMMENT 'Name shown. Can be changed to semantics: Important, Idea...',
+            background VARCHAR(25) NOT NULL COMMENT 'Any valid HTML way of representing color.',
+            border VARCHAR(25) NOT NULL COMMENT 'Any valid HTML way of representing color.',
+            is_usable TINYINT(1) NOT NULL DEFAULT '1',
```
fguillot: Just to follow the existing convention, can you rename `is_usable` to `is_active`?

app/Template/config/color.php
```diff
@@ -0,0 +1,36 @@
...
+        <tr>
+            <th><?= t('Color') ?></th>
+            <th><?= t('Label') ?></th>
+            <th><?= t('Background') ?></th>
+            <th><?= t('Border') ?></th>
+            <th><?= t('Selectable?') ?></th>
+        </tr>
+    </thead>
+    <tbody>
+    <? foreach ($colors as $color_id => $color) { ?>
```
fguillot: This syntax is not compatible with recent version of PHP, you should use something like:

```php
<?php foreach ($colors as $color_id => $color): ?>
...
<?php endforeach ?>
```

fguillot:
Thanks for submitting your pull-request and your patience. If you have time to fix the comments, I will be able to merge your pull-request and do more tests.

fguillot:
Closing, PR never updated.

PR #1412, opened 2015-10-26
Add filter by task link

oliviermaridat:
This PR add the possibility to filter tasks using task links. For example: status:open link:"is a milestone of".

I have added a TaskFilter::filterByLinkName method and used a similar logic as TaskFilter:filterBySubtaskAssignee. I have also updated english documentation and unit tests.

I did not understand the point of creating a TaskFilter::filterByLink method, (I did not find where TaskFilter::filterByColor, TaskFilter::filterByCategory, ... are used), so I did not create it yet. Please ask me if required.

### DIFF
--- a/app/Core/Lexer.php
+++ b/app/Core/Lexer.php
@@ -39,6 +39,7 @@ class Lexer
         "/^(swimlane:)/"                                 => 'T_SWIMLANE',
         "/^(ref:)/"                                      => 'T_REFERENCE',
         "/^(reference:)/"                                => 'T_REFERENCE',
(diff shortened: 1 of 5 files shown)


fguillot:
Can you update the unit test `Model\TaskFilterTest.php`?

I'm not sure if it's really pertinent to search by link id (ex: `link:42`)? usually people don't know database id, for tasks or projects ok but links...

oliviermaridat:
Yeah, it make sense, I will do it. Thanks.

oliviermaridat:
Test added. I discovered an issue (unknown link was filtering everything) that is now fixed.

PR #1416, opened 2015-10-26
Update TaskExport.php

Folcky:
Changed assignee_username to human name instead of username.

### DIFF
--- a/app/Model/TaskExport.php
+++ b/app/Model/TaskExport.php
@@ -57,7 +57,7 @@ class TaskExport extends Base
             tasks.color_id,
             tasks.date_due,
             creators.username AS creator_username,
-            users.username AS assignee_username,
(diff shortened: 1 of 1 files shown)


BlueTeck:
Maybe this should be configurable for Kanboard, also for the Board assignee name etc...

Folcky:
Why do you need username in the report? 
For example in transition export we can see that assignee name is a human name of an user.

I generate report every week and manually replace username with name :(. 
I dont understand for what purpose we need an option to configure that simple thing...

fguillot:
In this case it's probably better to add a new column in the export with the full assignee name.

Replacing the column username by the full name remove a feature that may necessary by other Kanboard users.

fguillot:
Now in master, the export show the assignee name and username.

PR #1440, opened 2015-11-01
build automation patch

phoen1x:
My goal is to spin up an Docker instance with Kanboard, do some automated tests and discard the Docker container.

The random API Token for the Json-RPC is a good security measure but essentially forbids build automation because my scripts can't obtain the API Token.

This could be a <http://en.wikipedia.org/wiki/Patch_(Unix)> build into the Dockerfile or a patch for the PHP source code. So please check for security issues before accepting.

##### More Information

http://github.com/phoen1x/skylar/blob/kanbaordBuildAutomation/startup.sh

<hr>
*CAUTION* the script below deletes the docker container named **kanboard**
<hr>

```bash
cd /tmp
git clone https://github.com/phoen1x/skylar.git
cd skylar
git checkout kanbaordBuildAutomation
./startup.sh
```

Thanks for providing such a awesome project!

Regards,
Thomas

### DIFF
--- a/app/Api/Auth.php
+++ b/app/Api/Auth.php
@@ -31,6 +31,8 @@ class Auth extends Base
             $this->userSession->refresh($this->user->getByUsername($username));
         } elseif ($username === 'jsonrpc' && $password === $this->config->get('api_token')) {
             $this->checkProcedurePermission(false, $method);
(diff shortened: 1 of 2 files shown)


fguillot:
I integrated your PR manually since I have done some code refactoring. Now, you can define the API token in the config file, that will override the one stored in the database.

Thanks.

PR #1459, opened 2015-11-10
Get informations from ldap directory even if we use SSO

lrrb:
If we enable SSO and Ldap in config.php, the first time a user ID is passed by the reverse proxy to Kanboard, informations about this user are bringing from the directory.

### DIFF
--- a/app/Auth/ReverseProxy.php
+++ b/app/Auth/ReverseProxy.php
@@ -66,7 +66,7 @@ class ReverseProxy extends Ldap
      */
     private function createUser($login)
     {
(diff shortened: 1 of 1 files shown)


lrrb:
Seems that updating line 70 with "if (LDAP_ACCOUNT_CREATION && is_array($this->lookup($login))) " would increase code quality.

fguillot:
Your code is not valid and it's breaking the unit tests.

By implementing this feature, your are changing the default behaviour of the ReverseProxy authentication method so that need to be documented and disabled by default for existing users of this feature.

Issue #1460, opened 2015-11-10
Url rewrite nginx documentation


filonenko-mikhail:
There are some errors in nginx config for url rewriting

param `try_files $uri $uri/ /index.php;` is incomplete
`fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;` is missed

Valid configuration on this date.

```
index index.php;

location / {
    try_files $uri $uri/ /index.php$is_args$args;

    # If Kanboard is under a subfolder
    # try_files $uri $uri/ /kanboard/index.php;
}

location ~ \.php$ {
    try_files $uri =404;
    fastcgi_split_path_info ^(.+\.php)(/.+)$;
    fastcgi_pass unix:/var/run/php5-fpm.sock;
    fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
    fastcgi_index index.php;
    include fastcgi_params;
}

# Deny access to the directory data
location ~* /data {
    deny all;
    return 404;
}

# Deny access to .htaccess
location ~ /\.ht {
    deny all;
    return 404;
}
```

fguillot:
The documentation have been updated. Thanks.

Issue #1352, opened 2015-10-11
Moving columns not possible if the rewrite is enabled


thelinuxfr:
Hello,

In the preferences of a project, I can not set the order of columns if the url rewrite is enabled with nginx anyway (/project/8/column/34/move/down?csrf_token=......).

I get an error message : Access forbidden

Thx

fguillot:
What Nginx config are you using?

thelinuxfr:
```
server {
        listen 80;
        server_name mydomain.com;
        return 301 https://$server_name$request_uri;  # enforce https

	access_log /var/log/nginx/kanboard.access.log;
	error_log /var/log/nginx/kanboard.error.log;

        location ~ \.php$ {
                fastcgi_pass    unix:/var/run/php5-fpm.sock;
                fastcgi_index   index.php;
                fastcgi_param   SCRIPT_FILENAME $document_root$fastcgi_script_na
me;
                include                 fastcgi_params;
        }

        location / {
                try_files $uri $uri/ index.php;
        }
}

server {
	listen		443 ssl spdy;
	server_name	mydomain.com;
	root		/var/www/kanboard/;
	index		index.php;

	access_log /var/log/nginx/kanboard.access.log;
	error_log /var/log/nginx/kanboard.error.log;

        ssl on;
	ssl_certificate /etc/nginx/certs/mydomain.com.crt;
	ssl_certificate_key /etc/nginx/certs/mydomain.com.key;

	location / {
    		try_files $uri $uri/ /index.php;
	}

	location ~ \.php$ {
    		try_files $uri =404;
    		fastcgi_split_path_info ^(.+\.php)(/.+)$;
   		fastcgi_pass unix:/var/run/php5-fpm.sock;
   		fastcgi_index index.php;
    		fastcgi_param   SCRIPT_FILENAME $document_root$fastcgi_script_na
me;
    		include fastcgi_params;
	}

	# Deny access to the directory data
	location ~* /data {
    		deny all;
    		return 404;
	}

	# Deny access to .htaccess
	location ~ /\.ht {
    		deny all;
    		return 404;
	}
}
```

perburn:
Can confirm this problem, also running nginx with the config suggested in the documentation.

perburn:
I was able to fix this issue by changing the 
try_files $uri $uri/ /index.php;
to
try_files $uri $uri/ /index.php$is_args$args;
as also suggested in issue https://github.com/fguillot/kanboard/issues/1460

[...]

PR #1525, opened 2015-12-04
Fix for #1352 based on #1460

perburn:
Fixes https://github.com/fguillot/kanboard/issues/1352 by implementing the changes suggested in https://github.com/fguillot/kanboard/issues/1460


fguillot:
Thanks.

PR #1548, opened 2015-12-14
Update Polish translation

dromek:
Hi 
I translated some phrases to polish lang (locale: pl_PL)

Regards
Damian


fguillot:
Thanks.

PR #1686, opened 2016-01-19
Add analysis divided for each swimlane in 'Task distribution'

Busfreak:
Hi,

I added the possibility to analyse the 'Task distribution' for each individual swimlane, if available.
Starting and standard behaviour is same as in your original analysis.

Added also count of closed tasks (number and percentage).

task distribution

Hope that helps :smile:


fguillot:
Your fork is probably too old or something else is wrong, the file `app/Model/ProjectAnalytic.php` have been removed recently.

Busfreak:
OOh, sorry, I will check this!

Issue #1705, opened 2016-01-22
Editing a task to move swimlanes


SchmilK:
I have a complaint from a user that has over a dozen swimlanes saying she can't easily move a task from say, swimlane 12 to swimlane 1 because it is off the screen, even when all swimlanes are minimized.

I noticed if you edit a task you can move it to a new project which has the ability to assign a swimlane, but simply editing a task does not have an entry to change swimlane.

BlueTeck:
Did you mean the same as in #696 ?

Busfreak:
I created a pull request resolving the missing field: https://github.com/fguillot/kanboard/pull/1707

SchmilK:
Yes thank you!

bugwatcher:
This issue has been closed automatically by a bot for inactivity. If you still need this feature, you should consider contributing to this open-source project. For more information, read the [FAQ](https://docs.kanboard.org/en/1.2.3/developer_guide/faq.html#why-do-you-close-inactive-issues-automatically).

PR #1707, opened 2016-01-22
 Editing a task to move swimlanes #1705 

Busfreak:
This pull request resolves issue https://github.com/fguillot/kanboard/issues/1705 and https://github.com/fguillot/kanboard/issues/696.

It adds a selector for swimlanes and the possibility to edit the estimated time in edit_task.


Busfreak:
updated for current codebase (1.0.25), so it's mergeable again

fguillot:
Adding the missing form element on the edit page is not enough. It's not available yet on the form because the task positions needs to be updated in the column source and destination.

Another approach might be to add a new page just to move a task by choosing the column/swimlane...

Busfreak:
OK, I will try a plugin.
THX

SchmilK:
A different user experienced the "how do I move a task to an item at the bottom of a swimlane?" today.  
Simply grabbing the task, then using the keyboard arrow keys scrolled the page down and he was able to  move it to a swimlane at the bottom of the screen.

PR #1723, opened 2016-01-25
Update task dropdown menu to make it work everywhere

oliviermaridat:
This PR updates the task dropdown menu (previously only available on the board view) to make it available on all views: dashboard, board, listing, Gantt and task link view.

The first commit updates the task dropdown menu behaviour in itself. The next commits make it available on each view.

If you have any question, please don't hesitate to ask me.

Notice:
* I didn't manage to make it work on Calendar view since HTML tags can't be used in the task title...
* If Gantt and Dashboard view where using the "show" action like the other controller, the code of Base.php and Url.php could be slightly simplified


fguillot:
fyi I started to simplify the page loading via ajax to avoid to pass a redirect variable everywhere. The dropdown menu will be there but I'm doing that step by step.

oliviermaridat:
Great! So basically the first commit will become useless, and the other ones may be used to add the task dropdown menu in the different views.

fguillot:
I think we have the dropdown almost everywhere now.

PR #1725, opened 2016-01-25
Update DE_de/translations.php

85pando:
Bran -> Surely a typo
Braun (Amber) -> Bernstein is actually the literal translation of amber and used for the color as well.
Lime -> With several shades of green, I think Giftgrün fits more than describing it with the english term.
Zuständig -> would be an adjective, Zuständiger is a person.


BlueTeck:
:+1:

fguillot:
Thanks!

Busfreak:
sorry, but lime is most likely "Lindgrün", alternatively "Limettengrün"

85pando:
Colors is always a difficult thing to name. Maybe we should open an Issue with color names, where we also have the actualy colors there, so we can discuss about the naming of them.

PR #1737, opened 2016-01-26
Update task link tooltip view

oliviermaridat:
This PR updates the task link tooltip (available on the board view) to make it clearer and to use the same "task progress" based on columns used on the Gantt chart.

Here is an example:

kanboard_tooltip


fguillot:
Nice!

PR #1739, opened 2016-01-26
Added indent and colon to russian translation

d4rk5eed:
- Average lead time in Analytics
- Average cycle time in Analytics

Before: 
prev

After:
new


fguillot:
Thanks!

PR #1741, opened 2016-01-27
Create an account without password and send an email to the new user

oliviermaridat:
This PR adds a new check-box in the "Create local user" form. If checked, an email is sent to the new user with a link to select a password (using the reset password mechanism). In this case, "password" fields became optionals (a random password of 12 characters is generated for the new user to prevent anyone to access the account before the real user) and "email" field became mandatory. 

This is useful for an administrator to create new user and invite them to join Kanboard. Before that, the administrator had to choose a password and to send it by email or post-it to the finale user.

If accepted, I have few enhancements in mind for the future:
* The import mechanism should also send an email to the new users when no password nor remote id is provided. But I don't know yet how to avoid code duplication for this use case...
* Add a "Application name" that will be displayed in the email instead of just "Kanboard" (e.g. "Kanboard CompanyG")
* Add a "state" (or similar) to the password reset mechanism in order to spot if it is an "invitation" or a "password reset request". This will allow to display a "Welcome" page to the new user instead of a "Password Reset" page.
* Does it make sense to do the same for remote user? To invite them to login with their Google or Github or ... account.
* Add a "Reset Password" button for administrator in a "Edit password" view of a user

All comments are welcome.

### DIFF
--- a/app/Controller/User.php
+++ b/app/Controller/User.php
@@ -114,6 +114,12 @@ class User extends Base
             $project_id = empty($values['project_id']) ? 0 : $values['project_id'];
             unset($values['project_id']);
 
+            $is_invitation = (isset($values['email_invitation']) && $values['email_invitation'] == 1);
+            if ($is_invitation) {
+                $chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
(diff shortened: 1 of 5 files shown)


fguillot:
It would be better to make a plugin for that.

fguillot:
Closing this PR. Should be implemented as plugin. However, missing plugin API can be added to implement this plugin correctly.

PR #1755, opened 2016-01-31
update chinese translation

renothing:
hi, sir
I completed 90% of chinese language. the rest will be finished later


fguillot:
Thanks!

PR #1774, opened 2016-02-03
Stale Apache pids make kanboard inaccessible

cstaylor:


fguillot:
The Docker image is using Nginx and php-fpm now.

Maybe you should try it again.

PR #1837, opened 2016-02-13
Fixed issue with customer filter popover and added confirmation for r…

nttict-nick-b:
Fixed up custom filter removal confirmation and blank popover as per issue 1825


fguillot:
Thanks, it's merged!

Issue #234, opened 2014-09-02
Custom Fields


timjcraig:
Ability to add custom fields to tasks, users and projects, also use custom fields in automatic actions.

timjcraig:
@fguillot are there any plans to add custom fields to roadmap?  Only feature restricting migration to Kanboard.

BlueTeck:
There is a new Plugin System in the next Version, so you could add anything you want.
https://github.com/fguillot/kanboard/blob/master/doc/plugins.markdown
But a little bit of coding is needed. 

I already thought about a plugin or a core function that allows you to add custom fields via GUI. Like the wordpress custom fields: https://codex.wordpress.org/Custom_Fields

fguillot:
@timjcraig I agree with the feature request but there is so much stuff to improve before, I will see how to implement that.

BlueTeck:
Technically it is now possible with the meta table: https://github.com/fguillot/kanboard/blob/master/doc/plugin-metadata.markdown

satisfiser:
THIS!!!  We are going to deploy kanboard in the next week or 2 but one of the items that would really help us would be the ability to add custom fields ... so here is another vote for the ability to have custom fields.

fguillot:
Basically, the feature is accessible programmatically for plugins but there is no user interface at the moment.

PR #1848, opened 2016-02-16
Hook for left column in task form

d4rk5eed:
Added hook 'template:task:form:right-column'
Could be useful for default form customization.


d4rk5eed:
relates to #234

PR #1868, opened 2016-02-19
template hook for css media = print

Busfreak:
I added a new hook in Template/layout (templare:layout:css_print) which calls the new function Helper->Asset->css_print. This function inserts a given css file as media='print'.

I also updated the documentation.

Related to https://github.com/fguillot/kanboard/issues/1864

### DIFF
--- a/app/Helper/Asset.php
+++ b/app/Helper/Asset.php
@@ -35,6 +35,19 @@ class Asset extends \Kanboard\Core\Base
     }
 
     /**
+     * Add a stylesheet asset media = print
+     *
+     * @param  string   $filename   Filename
+     * @param  boolean  $is_file    Add file timestamp
+     * @param  string   $media      Media
+     * @return string
+     */
+    public function css_print($filename, $is_file = true, $media = 'print')
+    {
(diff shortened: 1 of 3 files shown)


froschdesign:
> I added a new hook in Template/layout (templare:layout:css_print) which calls the new function Helper->Asset->css_print.

Sorry, but why is this needed? The 	third parameter in [`Kanboard\Helper\Asset::css`](https://github.com/fguillot/kanboard/blob/master/app/Helper/Asset.php#L32) gives you the possibility to set the value for the media attribute.
With your solution you must add 8 more methods for all other media types. (`all`, `aural`, `braille`, `handheld`, `projection`, `tty`, `tv`, `embossed` - `screen` and `print` is already there)

If you can not set the third method parameter for the asset helper, then the hook system is the problem and not the helper.

Busfreak:
Hello Frank,

I agree with you, but the hooking system isn't able to work with a third parameter. And changing the whole hooking system is a bit beyond of my possibilities.
If you could help changing the hooking system, I would be grateful.
Or do you know a different way?

Sincerely
Martin

PR #1890, opened 2016-02-25
Update es_ES translation

gerardozamudio:
translate missing strings


gerardozamudio:
Missed some double quotes on two %s, going to resubmit

PR #1941, opened 2016-03-04
Allow to use original template in overriden templates

oliviermaridat:


BlueTeck:
:+1:

oliviermaridat:
This PR adds the possibility to use the Kanboard original template inside a plugin template, even if it has been overridden. It updates the code and the doc.

The proposed syntax is "kanboard:folder/template" (which is similar to the "plugin:folder/template").

An example is available there: https://github.com/oliviermaridat/kanboard-milestone-plugin/blob/master/Template/tasklink/show.php#L13 

What do you think of this feature @fguillot? Do you approve the concept? Do you have other idea for the implementation?
Don't hesitate to ask me.

fguillot:
I merged manually your PR and changed a little bit the code.

Thanks.

PR #1942, opened 2016-03-04
Enhance task link view to mimic the subtask view

oliviermaridat:


oliviermaridat:
This update looks like this:

tasklinkview

PR #1963, opened 2016-03-07
Header redesign

mcorteel:
I introduced another `UrlHelper` method to create links with the `.btn` class automatically. This class parses the label to replace `<fa-icon>` with `<i class="fa fa-icon fa-fw"></i>`. This way, we don't add anything to the link helper (which is better for performance and compatibility).

I tried to make the change to all headers, but I might have missed some.

Anyway, I hope you like it!


fguillot:
Looks great but I'm not really fan of having a regex inside the button link helper. There is probably a better way of doing that. Maybe adding $icon as first parameter is more appropriate.

Not sure if we need to convert all header links to this design. What about starting only with view switcher and the filter dropdown?

mcorteel:
Yes, another parameter would be better. Also, maybe `button` would be a better name for the helper.
I like the idea of homogenous headers, but we can start with the view switchers if you want. What about the actions dropdown? I think it will look a little out of place if it's the only link in the header.

mcorteel:
OK, I made the changes. I pulled some of your commits when I reverted the changes to master state... I don't know if it makes a difference.

fguillot:
Thanks for updating your pull-request.

I was about to merge your branch even if there are some conflicts and I see that the buttons of dialog boxes are very small now. Is there a way to restore the original size?

Thanks

mcorteel:
I used the existing `.btn` class in the header for the sake of homogeneity. I think it's weird to have different button designs and it should be addressed globally.
I added a rule to adjust the font size and padding of buttons in `.form-actions` (to make them a little bigger).

fguillot:
I merged some of your changes.

Thanks.

PR #1966, opened 2016-03-07
Changing new self to new static, to accommodate subclasses

djpadz:
I created a subclass of Kanboard\Core\Ldap\Client, which overrides a handful of methods.  Using new static, instead of new self ensures that the new instance will be of the calling class, instead of Kanboard\Core\LDAP\Client.


fguillot:
Thanks Dj

Issue #1861, opened 2016-02-18
[FR][Metadata] add user and timestamp to metadata tables


BlueTeck:
Hi all,

it would be nice if the Metadata tables get a **timestamp and a user field**, filled with the user and date the entry was changed.

Regards,
Timo

Busfreak:
makes realy sense!
:+1:

BlueTeck:
PR merged

PR #1993, opened 2016-03-11
#1861 user and date for metadata

BlueTeck:
#1861 user and date for metadata

### DIFF
--- a/app/Model/Metadata.php
+++ b/app/Model/Metadata.php
@@ -81,19 +81,19 @@ abstract class Metadata extends Base
     public function save($entity_id, array $values)
     {
         $results = array();
-
+        $user_id = $this->userSession->getId();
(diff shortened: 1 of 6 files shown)


fguillot:
- The CI build doesn't pass, there is a typo in your unit test.
- Why do you need to have application metadata, is that not the same thing as the settings table?
- What do you think about extending the abstract class `Model\Metadata` to avoid code duplication?

Thanks for submitting a pull-request.

BlueTeck:
1. I will fix it
2. oh yes it is
3. I didn't use the class because you don't need the entity type/id for application settings

I will change 1 and 2, but what is your proposal about 3?

BlueTeck:
Changed it

BlueTeck:
Changed and rebased

PR #2001, opened 2016-03-13
Fix task age/time in column titles

mcorteel:
The "in days" description is wrong. I'm not sure if it was OK to do the `locale:sync` before the pull request.


fguillot:
Why is it wrong? the duration is shown in days?

Issue #2018, opened 2016-03-16
Activity stream design


mcorteel:
I have a design proposal for the activity stream. This is of course inspired by Github and issue #1995.
![spectacle tj3057](https://cloud.githubusercontent.com/assets/6164603/13809227/fe28c2b6-eb69-11e5-83fc-8f9f4562261f.png)
This is a partial implementation (it's functional though) but it shows how we could have a more compact activity stream. And with a more compact activity stream, it may be interesting to include it (as suggested in #1995) in the summary of the task. We could then integrate the other functions (sub-tasks/links/etc).

mcorteel:
Here is my design updated:
![spectacle b17175](https://cloud.githubusercontent.com/assets/6164603/13816901/75020a66-eb90-11e5-911e-7d26894c26d6.png)

BlueTeck:
The userimage should have the same size in every row.

mcorteel:
Like in github, I wanted to take advantage of events that don't need to display a content and make them smaller. This saves a lot of vertical space, so that we can display more activity.
If other users feel like you, we can just as easily have a fixed avatar size.

mcorteel:
It's not that bad with a 32px avatar for everything:
![spectacle c20622](https://cloud.githubusercontent.com/assets/6164603/13818601/d163b9f2-eb96-11e5-9552-518ba0df96f4.png)
It does look more natural.

fguillot:
Closing, see comments in PR

Issue #126, opened 2014-06-20
Avatar for users


biazouk:
Profile picture or avatar for user.
Image on tasks rather than name, with name label when you hover over the image.

nekohayo:
There is also the potential for this to create a more compact design, by replacing the full-length name labels in the board view by icons that contain either the (gr)avatar or, if no image is available, the first letter of the firstname with a unique color, like Google Documents does:

![foo](https://cloud.githubusercontent.com/assets/479401/8356875/df296be6-1b25-11e5-9785-fb3928f6812c.png)

This would make it future-proof for a case where a task can have multiple people assigned/co-assigned to it, as per issue #85. Of course, the image above is just an example of the "fake avatars with letters" concept, the layout of them is intended to be much more compact (in a line) on the board view, like Trello does:

![home-hero_es](https://cloud.githubusercontent.com/assets/479401/8356955/484a8d76-1b26-11e5-83e9-cd20071d5f37.png)

NinjaPewPew:
I hope this will come true :+1:

tk-wfischer:
I'd also love this feature :-)

cmitasch:
:thumbsup:

mcorteel:
Gravatar also provides Identicons like this:
![Identicon](http://www.gravatar.com/avatar/3b3be63a4c2aba9b013787725dfce802?d=identicon)
This would be really easy to implement and this way, each user would have a unique avatar when gravatar integration is active. We could then integrate more avatars. @fguillot  what do you think?

nekohayo:
@mcorteel well, in terms of informativeness those gravatar identicons are useless compared to the letter-based ones; and, on a sidenote, I really cannot expect my users to _ever_ bother setting up a gravatar account (vs me uploading a profile picture for them in the kanboard admin interface), they have zero interest in technology.

That said, "gravatar for real avatars + fallback to letter-based identicons à la Google" would be a reasonable compromise.

fguillot:
All right, I think it's time to have avatars in Kanboard.

For a first iteration, having the possibility to upload an image and fallback to a letter-based identicon would be a good start. Letter-based avatars are the default on Google and Apple products and I assume they are better UX designer than me :)

Now the big question is where to place the avatar on the card? the bottom left seems to be the more appropriate.

mcorteel:
Something like that?
[...]

PR #2019, opened 2016-03-16
Activity redesign

mcorteel:
See #2018.


mcorteel:
I also adapted the comments for better integration.

froschdesign:
And what we see without Gravatar?

mcorteel:
There is nothing when the integration is disabled:
spectacle x15974
If the integration is enabled, the avatars are displayed:
spectacle h15974

froschdesign:
@mcorteel 
> There is nothing when the integration is disabled

Bad! :-1: 

You solution needs an external service and is restricted to a Gravatar.

[The first letter of the name is a good idea.](https://github.com/fguillot/kanboard/issues/126#issuecomment-115278666) This should be the standard implementation.
Gravatar or other services should be integrated via Plugin.

mcorteel:
This is just a redesign, I haven't changed anything else.
This follows the current avatar principle (if gravatar integration is enabled, it displays the avatar if available).
Displaying something when gravatar is disabled would be another pull request (I plan on making a proposal if this is merged, but it has to be merged first).

froschdesign:
> This is just a redesign, I haven't changed anything else.

Then we need to rethink here

mcorteel:
Please comment on issue #126 for that.

froschdesign:
A deeper integration of the Gravatar service in the core is the wrong way! (imo)

And "There is nothing when the integration is disabled" is not a nice idea.

mcorteel:
This is not the subject of this pull request, please comment on #126.

froschdesign:
> This is not the subject of this pull request…

"Show a Gravatar image or nothing" is not the topic?!

mcorteel:
No, it's not. This is the topic of #126, which will in turn be the topic of **another** pull request (if and when this one is merged). I haven't changed the behavior here.
Basically, you are telling the painter that you want another socket to plug in your toaster when he's asking you about the color of the walls. Please wait for the meeting with the electrician to talk about that.
If it helps, I do agree with you. It just isn't the right place to discuss this.

mcorteel:
@fguillot I see that you are partially redesigning the activity feed with the new (awesome) avatar feature. Don't you want to consider this pull request?

Kolesar:
Can we group activities?
I made few changes about this. Make a pagination, searchable and order by activity ID (it is by date) and order by Task ID, but from my perspective better way (exactly I want to see that solution :) ) is the grouping by date and by task in expandable list.

fguillot:
@mcorteel Thanks for submitting a pull-request and sorry about the delay of my answer.

As you can see, I also updated the activity stream to display the default avatar. Actually, that looks like that in master:

mon_flux_d_activite

Since everybody have different personal taste about the design, I prefer to keep it simple as it right now. 

But, if you prefer your own design, you still have the possibility to make a "theme plugin". I will be happy to publish it on the website.

Thanks.

mcorteel:
Thanks, I will look into it!

PR #2049, opened 2016-03-25
Task Footer Icon Hook

BlueTeck:
added task footer icon hook to add new hover/tooltip icon


fguillot:
Can you rename the hook to `template:board:task:icons` instead of `footer_icon`? thanks

BlueTeck:
Renamed and rebased :)

PR #2066, opened 2016-03-29
Generate avatar color based on name instead of initials

### LINKED ISSUE #2062: Option to disable (or change?) avatars
### Actual behaviour

I want to identify who a given task belongs to. However, we have two people in the organisation with the initials "SB". Because we use LDAP authentication, it uses the system default of SURNAME FIRSTNAME, so both default avatars appear as "BS", which isn't either of their names (and even it if were, it would be impossible to tell which one was which!)
### Expected behaviour

It'd be nice to be able to customise what someone's initials are, so for the two SBs, they can be labeled "SB1" and "SB2". Or just disable them, so that it just shows the name. I don't mind - either/or. But as it is, the tasks are very difficult to identigy
### Configuration

Kanboard v1.0.27

mcorteel:
Fixes #2062.

### DIFF
--- a/app/User/Avatar/LetterAvatarProvider.php
+++ b/app/User/Avatar/LetterAvatarProvider.php
@@ -31,7 +31,7 @@ class LetterAvatarProvider extends Base implements AvatarProviderInterface
     public function render(array $user, $size)
     {
         $initials = $this->helper->user->getInitials($user['name'] ?: $user['username']);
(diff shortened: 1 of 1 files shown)


fguillot:
Don't forget to check the unit tests. The build was broken.

mcorteel:
Sorry!

erpnedir:
Hi, how can I use this feature?

PR #2129, opened 2016-04-14
Fixing some pathing information for config.php and sqlite DB

djpadz:
We have the use case, where our Kanboard source tree is outside of the docroot.  As a result, our docroot contains only the following:

```
lrwxr-xr-x  1 www  www  13 Apr 13 17:39 assets@ -> ../git/assets
lrwxr-xr-x  1 www  www  16 Apr 13 17:39 index.php@ -> ../git/index.php
lrwxr-xr-x  1 www  www  18 Apr 13 17:39 jsonrpc.php@ -> ../git/jsonrpc.php
```

Prior to this patch, there also had to be a symlink for `data`, as well as the presence of a `config.php`.  With this patch, these resources can live in the source tree directory.  Check my math, however, and make sure that it won't break another use case.

### DIFF
--- a/app/common.php
+++ b/app/common.php
@@ -14,12 +14,12 @@ if (getenv('DATABASE_URL')) {
     define('DB_NAME', ltrim($dbopts["path"], '/'));
 }
 
-if (file_exists('config.php')) {
-    require 'config.php';
+if (file_exists(__DIR__.DIRECTORY_SEPARATOR.'config.php')) {
+    require __DIR__.DIRECTORY_SEPARATOR.'config.php';
 }
 
-if (file_exists('data'.DIRECTORY_SEPARATOR.'config.php')) {
-    require 'data'.DIRECTORY_SEPARATOR.'config.php';
+if (file_exists(__DIR__.DIRECTORY_SEPARATOR.'..'.DIRECTORY_SEPARATOR.'data'.DIRECTORY_SEPARATOR.'config.php')) {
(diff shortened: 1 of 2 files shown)


fguillot:
I made a couple of improvements after merging your pull-request.

I also updated the installation instructions for people who needs a similar setup: https://github.com/fguillot/kanboard/blob/master/doc/installation.markdown#installation-outside-of-the-document-root

Thanks

app/common.php
```diff
@@ -14,12 +14,12 @@
     define('DB_NAME', ltrim($dbopts["path"], '/'));
 }
 
-if (file_exists('config.php')) {
-    require 'config.php';
+if (file_exists(__DIR__.DIRECTORY_SEPARATOR.'config.php')) {
+    require __DIR__.DIRECTORY_SEPARATOR.'config.php';
```
fguillot: To avoid a regression, the config file have to be loaded from the root directory of Kanboard, not from `app/config.php` (`..` is missing).

(fixed in commits after yours)

djpadz: I think I started there, but __DIR__ was coming back as being /path/to/kanboard/app.  That's why I included the "..".  Remember, the change is being made to app/common.php, in which __DIR__ would be app.

djpadz: I see what you did there...  For a minute, I thought I had the `..` in there, and you'd removed it.  Thanks for adding it!

PR #2144, opened 2016-04-20
Added group by date/task overdue tasks on notification page

Kolesar:
Added:
- Group by date
- Group by task
on notification page

*Notification:*
Lookup maybe is not so good, but I tried to keep simple as much as possible


fguillot:
Closing this PR because it's not mergeable anymore. You should put the logic that groups tasks in a specific method instead of changing the existing method. Try to send only files necessary to the PR, it's easier to review.

Thanks

PR #2147, opened 2016-04-20
Removal of min-height

Cybergern:
This removes setting minimum height on swimlanes. It seems to have been
previously necessary because of a dropdown menu but no longer seems
necessary since the current dropdown menu floats on top of the swimlane.


fguillot:
Your pull-request seems to introduce a regression that prevent the user to drag a task a the bottom of the columns.

Cybergern:
Ahh, you are correct, I did not realize that was a consequence of my change. Obviously a more refined change needs to be made instead.

PR #2170, opened 2016-04-24
Button links

mcorteel:
As promised, I'm submitting a pull request to include icons in link tags. I changed the `UrlHelper::button` method to include all `UrlHelper::link` parameters for compatibility.
This would make theming a lot easier.
Note: unit tests fail, but they fail on master too...


fguillot:
I don't get why that simplify theme creation?

mcorteel:
Because you cannot theme `<i></i><a></a>` as a button properly, but you can do that with `<a><i></i></a>`!
Anyway, you're doing a great job with kanboard but I'm not using it anymore (I use the simple boards provided by phabricator instead) so I won't be bothering you with my ideas anymore :).

fguillot:
All right. I close this PR because I cannot merge it anymore (sorry to review so lately).

Issue #2258, opened 2016-05-22
1.0.28 - Remove task: Page not found -- Sorry, I didn't find this information in my database! 


jtourt:
### Expected behaviour

After removing a task, browser should redirect to board view.

### Actual behaviour

Page not found 
Sorry, I didn't find this information in my database! 

![image](https://cloud.githubusercontent.com/assets/14179005/15454251/35327254-2001-11e6-9b21-e1e84a15bf3e.png)

### Steps to reproduce

1. Open any task.
2. Click Remove.
3. Click Yes to confirm that you really want to remove this task.

### Configuration

Application version: 1.0.28 
◾ PHP version: 5.6.20-0+deb8u1 
◾ PHP SAPI: fpm-fcgi 
◾ OS version: Linux 4.1.19-v7+ 
◾ Database driver: mysql 
◾ Database version: 5.5.44-0+deb8u1 
◾ Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2486.0 Safari/537.36 Edge/13.10586

fguillot:
This issue is now fixed in the development branch. Thanks.

Duplicate with #2201

PR #2259, opened 2016-05-22
Fix Page not found error

jtourt:
Resolves issue #2258.


fguillot:
You can remove a task from multiple places in the user interface with the dropdown menu. With your modification, after removing a task the user is always redirected to the board. The user should be redirected to the board only if he is already on the board view or in the task view.

PR #2271, opened 2016-05-29
Fix task date should not change when deplaced in the same column

### LINKED ISSUE #2262: "Days in this column" value should not be changed when a task moved in the same column
### Expected behaviour

"Days in this column" value should not be changed when a task moved to down or up in the same column.

### Actual behaviour

I have 2 tasks in WIP column. After I move first task under of second task, "days in this column" value becomes < 15m.

### Steps to reproduce

1. Create 2 tasks in WIP column
2. Wait 1 hour.
3. Change order of tasks. 

### Configuration

◾ Application version: 1.0.28 
◾ PHP version: 7.0.5 
◾ PHP SAPI: apache2handler 
◾ OS version: Windows NT 6.3 
◾ Database driver: postgres 
◾ Database version: 9.5.2 
◾ Browser: Mozilla/4.0 (compatible; MSIE 7.0; Windows NT 6.1; WOW64; Trident/7.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729; Media Center PC 6.0; .NET4.0C; .NET4.0E; GWX:MANAGED)

kerro:
fixes #2262

### DIFF
--- a/app/Model/TaskPositionModel.php
+++ b/app/Model/TaskPositionModel.php
@@ -142,6 +142,13 @@ class TaskPositionModel extends Base
             ->asc('id')
             ->findAllByColumn('id');
 
+        $task = $this->taskFinderModel->getById($task_id);
+
+        $is_same_column = false;
+        if (isset($task['column_id']) && $task['column_id'] === $column_id) {
+            $is_same_column = true;
+        }
+
         $offset = 1;
 
         foreach ($tasks_ids as $current_task_id) {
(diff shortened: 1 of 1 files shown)


kerro:
Is it ok to add a boolean argument `$update_date_moved` to the function `TaskPositionModel::saveTaskPositions` ?

`saveTaskPositions($project_id, $task_id, $position, $column_id, $swimlane_id, $update_date_moved)`

I don't like to add this particular case but I don't see any other way with the current implementation.

fguillot:
Thanks for submitting a pull-request. The code have been modified in the master branch.

Sorry for answering so lately.

PR #2302, opened 2016-06-06
Fix plural form

levlaz:
When removing all columns from a board, an alert box is displayed that
states "Your board doesn't have any column!", the correct plural form
should be "Your board doesn't have any columns!"


fguillot:
Thanks

PR #2303, opened 2016-06-06
Fix li element list-style-type

levlaz:
The tooltip was showing a bullet point for an `li` element which was showing up on the outside of the tooltip box. Setting the list-style-type to none removes this bullet point and makes the tooltip look a lot cleaner.

### Before 
screenshot 2016-06-06 14 26 08

### After
screenshot 2016-06-06 14 52 33

You can easily confirm that this has the expected behavior by adding the snippet below to the custom stylesheet under application settings. 

```
.ui-tooltip li {
  list-style-type: none;
}
```

### DIFF
--- a/assets/css/src/tooltip.css
+++ b/assets/css/src/tooltip.css
@@ -68,6 +68,10 @@ div.ui-tooltip {
     margin-bottom: 0px;
 }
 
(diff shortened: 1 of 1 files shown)


fguillot:
Thanks

PR #2312, opened 2016-06-07
Expose SubTask Time Tracking though the User API

ngtech:
We are using those in a desktop time tracking applet that needs to be able to log time spent in subtasks and to also create new subtasks


app/ServiceProvider/ApiProvider.php
```diff
@@ -60,6 +61,7 @@ public function register(Container $container)
             ->withObject(new ProjectApi($container))
             ->withObject(new ProjectPermissionApi($container))
             ->withObject(new SubtaskApi($container))
+            ->withObject(new SubtaskTimeTracking($container))
```
fguillot: The class name should have the suffix "Api" like the other ones to be consistent.

ngtech:
I renamed the class , adding the "Api" suffix , making it consistent with the other API classes. 

By the way many thanks for making Kanboard - it has significantly improved our everyday operations.

angystardust:
Hi @ngtech , may I know which is the desktop time tracking software you are using?

ngtech:
It's an electron "kanboard client" app I wrote and I am still developing - it's currently in a very early / alpha stage of development without a decent user interface but  I am planing to release it once it hits a useful level of functionality.

PR #2328, opened 2016-06-09
Add column dropdown to edit task

### LINKED ISSUE #1879: moving task without drag and drop
we are in a test phase so far we very happy with kanboard. I have around 34 columns going in a manufacturing facility with over 200 tasks active with 20 users.  problem I get is moving a task to columns off the page.  is there a way to move a task across columns without dragging?  similar action to changing assignee goes to pulldown menu but for columns?

levlaz:
This allows users to move a task to a different column without having
to drag it.

fix #1879


fguillot:
Your PR doesn't change the task position (column position of the table tasks). If you edit a task and choose another column with existing tasks you will end up with tasks with the same position.

levlaz:
Oh I'm sorry .. I misread that. Yes I will address keeping the same position.

levlaz:
When I test this locally it works for me. :/ 

What's missing?

PR #2378, opened 2016-06-22
Hide tasks within specific columns in dashboard

Busfreak:
If you have a column in your board, lets say backlog, with tasks not started yet, you can hide the tasks in this column from the dashboard overview with this PR.

The edit and create dialog for columns have a new checkbox:
edit

In the project summary you can see the status of the columns:
summary

In the dashboard the task from column backlog is not shown:
dashboard

But in the board view you can see it:
board

### DIFF
--- a/app/Controller/ColumnController.php
+++ b/app/Controller/ColumnController.php
@@ -66,7 +66,7 @@ class ColumnController extends BaseController
         list($valid, $errors) = $this->columnValidator->validateCreation($values);
 
         if ($valid) {
-            if ($this->columnModel->create($project['id'], $values['title'], $values['task_limit'], $values['description']) !== false) {
+            if ($this->columnModel->create($project['id'], $values['title'], $values['task_limit'], $values['description'], $values['hide_in_dashboard']) !== false) {
                 $this->flash->success(t('Column created successfully.'));
                 return $this->response->redirect($this->helper->url->to('ColumnController', 'index', array('project_id' => $project['id'])), true);
             } else {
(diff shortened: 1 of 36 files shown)


Busfreak:
Hi @fguillot !
This PR runs in merge conflicts now. Shall I resolve them?

fguillot:
Why do you need this feature?

Busfreak:
After some events, lets say external audits, I create lots of tasks (50-80) in bulk, to work with the different submissions. This happens twice a year, but the work has to be spread over the whole year. If I want a task to be started, I assign it to someone and move it to the next (to be done) column. In the dashboard (my tasks) I then have lots of tasks, which are not started, and I can't focus on the work which has to be done now. This feature makes it very much easiser to keep the overview!

macman31:
@Busfreak your use case would very likely be resolved in a better way by #2389 i.e. having the ability to create unassigned tasks instead of having those not-yet-assigned tasks assigned to yourself and hidden on the board.

Busfreak:
@macman31 
You are only partial right, because some tasks will be assigned to someone (me or someone else) when I know who will have to work with it. If I don't assign them I have to write it in the comment or so and assign it later, when it has to start. That's no real solution.

macman31:
@Busfreak I understand your need now, and I think that your solution is indeed a good one, maybe in addition to a "create as unassigned" option!

Busfreak:
The conflicts have been resolved in a new PR: https://github.com/fguillot/kanboard/pull/2424

PR #2388, opened 2016-06-26
New action TaskCloseNoActivityColumn

tester22:
Based on TaskCloseNoActivity but with the changes that only tasks that are in the defined column will be closed.
We use this to clean our done column when task havn't been changed for two weeks.


app/Action/TaskCloseNoActivityColumn.php
```diff
@@ -0,0 +1,98 @@
...
+     * @return bool            True if the action was executed or false when not executed
+     */
+    public function doAction(array $data)
+    {
+        $results = array();
+        $max = $this->getParam('duration') * 86400;
+
+        foreach ($data['tasks'] as $task) {
+            $duration = time() - $task['date_modification'];
+			//echo $task['column_id'];
```
fguillot: You should remove these useless comments.

tester22:
The commented sections are removed.

PR #2389, opened 2016-06-26
Allow Unassigned task creation in team projects with one active user …

yakovenkov:
I create some team projects and start to creating tasks to see the whole picture.
Users are not assigned to projects yet, so all of the tasks are assigned to me because of missing Unassigned option in assignee selection list.

This makes difficult to distinguish between these newly created tasks and tasks that are actually assigned.

I see a parameter for `getAssignableUsersList()` function named `single_user.`
I think it would be better to allow Unassigned option for Team projects.


yakovenkov:
Sorry, it seems that rebasing pull-request code was a wrong decision %)

PR #2421, opened 2016-07-03
Fixed "due today" highlighting

kiney:
The highlighting for "due date" = today ignores the month and year.
So e.g. a task due on 03.08.2016 is also colored blue today (03.07.2016).
I considered this a bug and changed the behaviour.


angystardust:
@kiney I've also noticed that on my setup and i was going to submit an issue. Thanks for the proactive fix ;)

fguillot:
Thanks!

PR #2424, opened 2016-07-04
Hide tasks within specific columns in dashboard

Busfreak:
If you have a column in your board, lets say backlog, with tasks not
started yet, you can hide the tasks in this column from the dashboard
overview with this PR.
For details see https://github.com/fguillot/kanboard/pull/2378


app/Schema/Postgres.php
```diff
@@ -6,7 +6,12 @@
 use Kanboard\Core\Security\Token;
 use Kanboard\Core\Security\Role;
 
-const VERSION = 90;
+const VERSION = 91;
+
+function version_91(PDO $pdo)
+{
+    $pdo->exec("ALTER TABLE columns ADD COLUMN hide_in_dashboard INTEGER DEFAULT 0 NOT NULL");
```
fguillot: For Postgres, it's probably better to use the type `BOOLEAN` instead (same as other flags). If you can make this change, I'm good to merge.

Busfreak:
@fguillot 

I changed it to boolean.

Thanks a lot!!

PR #2431, opened 2016-07-04
Allow Unassigned task creation in team projects with one active user …

yakovenkov:
I create some team projects and start to creating tasks to see the whole picture.
Users are not assigned to projects yet, so all of the tasks are assigned to me because of missing Unassigned option in assignee selection list.

This makes difficult to distinguish between these newly created tasks and tasks that are actually assigned.

I see a parameter for getAssignableUsersList() function named single_user.
I think it would be better to allow Unassigned option for Team projects.

### DIFF
--- a/app/Controller/TaskBulkController.php
+++ b/app/Controller/TaskBulkController.php
@@ -32,7 +32,7 @@ class TaskBulkController extends BaseController
             'project' => $project,
             'values' => $values,
             'errors' => $errors,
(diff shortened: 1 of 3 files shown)


fguillot:
Sorry about the very long delay, your PR is finally merged.

yakovenkov:
Thank you! It's THE really great software!

30 Янв 2017 г. 4:16 пользователь "Frédéric Guillot" <
notifications@github.com> написал:

PR #2435, opened 2016-07-05
The generated QR Code for 2FA now contains an issuer.

stinnux:
The issuer can be configured in the config.php as SERVER_NAME and defaults to $_GLOBALS['SERVER_NAME']


app/Auth/TotpAuth.php
```diff
@@ -117,13 +117,13 @@ public function getSecret()
      * @param  string $label
      * @return string
      */
-    public function getQrCodeUrl($label)
+    public function getQrCodeUrl($label, $options)
```
fguillot: If you add a new argument, you should update the phpdoc of the method and specify the type if not scalar (array).

app/Auth/TotpAuth.php
```diff
@@ -117,13 +117,13 @@ public function getSecret()
...
     {
         if (empty($this->secret)) {
             return '';
         }
 
-        return GoogleAuthenticator::getQrCodeUrl('totp', $label, $this->secret);
+        return GoogleAuthenticator::getQrCodeUrl('totp', $label, $this->secret,false, $options);
```
fguillot: According to the signature of the method `GoogleAuthenticator::getQrCodeUrl()` the fourth argument is an integer not a boolean, the default is null.

app/Controller/TwoFactorController.php
```diff
@@ -56,6 +56,10 @@ public function show()
         $label = $user['email'] ?: $user['username'];
         $provider = $this->authenticationManager->getPostAuthenticationProvider();
 
+        print "SN: " . SERVER_NAME."<br>";
```
fguillot: Why do print stuff in the middle of the method?

app/constants.php
```diff
@@ -134,3 +134,5 @@
 defined('HTTP_PROXY_USERNAME') or define('HTTP_PROXY_USERNAME', '');
 defined('HTTP_PROXY_PASSWORD') or define('HTTP_PROXY_PASSWORD', '');
 defined('HTTP_VERIFY_SSL_CERTIFICATE') or define('HTTP_VERIFY_SSL_CERTIFICATE', true);
+
+defined('SERVER_NAME') or define('SERVER_NAME', $_SERVER['SERVER_NAME']);
```
fguillot: The name of this parameter should be something else because you use that generate an issuer for the TOTP.

You should also check for the existence of the key `SERVER_NAME`, it's not always available, that will generate a PHP notice in certain conditions.

fguillot:
Thanks for submitting a pull-request, it would be nice to address the comments and make sur the automated tests pass, they are currently broken.

stinnux:
I think i've addressed all issues you had. 

Accidently another change was added to the pull request (https://github.com/fguillot/kanboard/pull/2435/commits/035c6225b9b4db6765a60d475b676b5c739b23a7).

I don't know how to remove it from the pull request, because this is a different feature.

fguillot:
Your PR have been merged manually (cherry-pick the commits and squashed into a single commit).

You should make a branch for each PR to avoid unrelated commits.

Thanks.

PR #2539, opened 2016-08-01
Allow priority changes for inverted priority scales

rys:
It's common to want an 'inverted' priority range, for example where P0 is the highest. The task helper logic would hide priority editing when `priority_end` was larger than `priority_start`. This logic change allows it by checking that `priority_start` and `priority_end` are different instead.


rys:
I forgot to mention in the pull request that I've run the test suite and all tests pass.

fguillot:
Thanks!

PR #2586, opened 2016-08-16
Added template hooks for menu on project list

stinnux:
Added hooks for the menu of the project list (before and after the existing menu entries).


fguillot:
Thanks

PR #2617, opened 2016-08-29
Hook in ColorModel->getList

Busfreak:
Hello,

I need to do changes in the ColorModel default-colors list.
Here is a hook for changing color labels in ColorModel->getList.

Thanks for merging :smile:


fguillot:
Instead of having a condition, you should use the new "reference" hooks: `$this->hook->reference('my-hook', $listing);`. This is more lightweight...

In your plugin you can alter the variable `$listing` like that:

```php
$this->hook->on('my-hook', function(array &colors) {
  // do something with $colors..
  $colors['pineapple'] = 'Pineapple';
});
```

For the hook name, `model:color:get-list` should be enough. And do not forget to update the documentation as well. Thanks!

Busfreak:
Hi @fguillot 

thanks for the hint! I will do it.
But I have a problem getting the result back into ColorModel->getList.

Debuging the function Core/Hook->reference shows that the params are transferred correctly to my function, but the result isn't transferred back from function Core/Hook->reference, but the original $param. Do we need someting like
"$param = $listener($param);" instead of "$listener($param);"
in line 111 in the reference function?

Regards
Martin

Busfreak:
@fguillot 

Sorry, I got it now! There is a small but important "&" I neglected.

Can you please merge now?

app/Helper/FormHelper.php
```diff
@@ -174,7 +174,7 @@ public function textarea($name, $values = array(), array $errors = array(), arra
 
         $html = '<textarea name="'.$name.'" id="form-'.$name.'" class="'.$class.'" ';
         $html .= implode(' ', $attributes).'>';
-        $html .= isset($values->$name) ? $this->helper->text->e($values->$name) : isset($values[$name]) ? $values[$name] : '';
+        $html .= isset($values[$name]) ? $this->helper->text->e($values[$name]) : '';
```
fguillot: This change should not be part of your PR.

app/Model/ColorModel.php
```diff
@@ -154,6 +154,15 @@ public function getList($prepend = false)
             $listing[$color_id] = t($color['name']);
         }
 
+        $hook = 'model:color:modification:get-list';
+        if ($this->hook->exists($hook)) {
```
fguillot: This line and the condition below is not necessary anymore.

fguillot:
Can you address the inline comments and I will merge your PR? thanks.

PR #2474, opened 2016-07-13
Allow multiple boards on one page

stinnux:
I created a plugin (https://github.com/stinnux/kanboard-bigboard) that allows displaying multiple kanboards/projects on a single page. 

To enable the plugin various functions need to be changed, so that the boardId/projectId can be passed on. 

Also Added Hooks for project list menu

PR #2624, opened 2016-08-31
Feature multiboard

stinnux:
These PR allows having multiple Boards (for different projects) on one page. Basically all it does is forward the project id/board id to the relevant functions.

This is necessary for the Plugin https://github.com/stinnux/kanboard-bigboard. 

This replaces PR #2474


fguillot:
I found a regression, the keyboard shortcuts don't seems to work anymore (collapse tasks).

stinnux:
Where are the keyboard shortcuts defined? Can't find it.

aljawaid:
This would be a very good feature

orware:
I agree this would be a great feature (from the rough description above I think it's similar to an idea I was wanting to have back after we started using Kanboard earlier this year).

ArmReu:
Any update on this feature? We'd absolutely would love to use it when release!

PR #2691, opened 2016-09-21
Fix translations to Russian for docs and localization, grammarnaci etc

vistar:


fguillot:
I merged the commit about the translation. Other commits are the same as the other pull-request.

Thanks

PR #2707, opened 2016-09-28
Updated hooks to allow plugins interaction with subtasks

eSkiSo:
Moved plugin added fields on the first column to appear between tag selection and checkbox.
+
Added hooks to subtask operations to allow plugin interaction (Already have plugin that adds due date to subtasks but need this updates in order to work and display correctly)

### DIFF
--- a/app/Template/dashboard/subtasks.php
+++ b/app/Template/dashboard/subtasks.php
@@ -10,7 +10,7 @@
             <th class="column-20"><?= $paginator->order(t('Project'), 'project_name') ?></th>
             <th><?= $paginator->order(t('Task'), 'task_name') ?></th>
             <th><?= $paginator->order(t('Subtask'), \Kanboard\Model\SubtaskModel::TABLE.'.title') ?></th>
-            <?= $this->hook->render('template:dashboard:subtasks:header', array('values' => $values, 'errors' => $errors, 'paginator' => $paginator)) ?>
+            <?= $this->hook->render('template:dashboard:subtasks:header', array('paginator' => $paginator)) ?>
             <th class="column-20"><?= t('Time tracking') ?></th>
         </tr>
         <?php foreach ($paginator->getCollection() as $subtask): ?>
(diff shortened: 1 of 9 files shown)


app/Template/dashboard/subtasks.php
```diff
@@ -10,6 +10,7 @@
             <th class="column-20"><?= $paginator->order(t('Project'), 'project_name') ?></th>
             <th><?= $paginator->order(t('Task'), 'task_name') ?></th>
             <th><?= $paginator->order(t('Subtask'), \Kanboard\Model\SubtaskModel::TABLE.'.title') ?></th>
+            <?= $this->hook->render('template:dashboard:subtasks:header', array('values' => $values, 'errors' => $errors, 'paginator' => $paginator)) ?>
```
potorange: tested this patch against 1.33, $values and $errors are undefined at this place

app/Template/subtask/table.php
```diff
@@ -7,6 +7,7 @@ class="subtasks-table table-striped table-scrolling"
         <tr>
             <th class="column-40"><?= t('Title') ?></th>
             <th><?= t('Assignee') ?></th>
+            <?= $this->hook->render('template:subtask:table:header', array('values' => $values, 'errors' => $errors)) ?>
```
potorange: again $values and $errors are undefined at this place, replace with 'subtask' => $subtask may be ?

app/Template/dashboard/subtasks.php
```diff
@@ -10,6 +10,7 @@
             <th class="column-20"><?= $paginator->order(t('Project'), 'project_name') ?></th>
             <th><?= $paginator->order(t('Task'), 'task_name') ?></th>
             <th><?= $paginator->order(t('Subtask'), \Kanboard\Model\SubtaskModel::TABLE.'.title') ?></th>
+            <?= $this->hook->render('template:dashboard:subtasks:header', array('paginator' => $paginator)) ?>
```
fguillot: The name of this hook should be more specific because this is included between specific columns.

app/Template/subtask/table.php
```diff
@@ -7,6 +7,7 @@ class="subtasks-table table-striped table-scrolling"
         <tr>
             <th class="column-40"><?= t('Title') ?></th>
             <th><?= t('Assignee') ?></th>
+            <?= $this->hook->render('template:subtask:table:header', array()) ?>
```
fguillot: You can remove the 2nd argument if you pass nothing.

fguillot:
Thanks!

PR #2717, opened 2016-09-30
Update RU translations.php

dmkcv:


acs-ferreira:
Already included in #2705

fguillot:
This PR is in conflicts with #2717 which is merged. I don't speak Russian, so I cannot tell which one have the better translation.

PR #2725, opened 2016-10-05
Corrected translation.

eSkiSo:
Before to submit your pull-request:

- Be sure that the unit tests pass
- If you create a new feature, test your code, do not introduce new bugs
- Avoid code duplication
- Small pull-requests are easier to review and can be merged quickly
- 1 pull-request == 1 feature/improvement
- Non necessary features should be implemented as plugin


Shaxine:
Thanks for the correction of some of mine misspelling (Sublime doesn't have spellcheck) and I should be more focused on the task next time.
But don't you think using "Função" instead of "Cargo" gets a little misleading?

PR #2766, opened 2016-10-19
Search with 2+ arguments using double quotes is now working

AlanRezende:
The searching with 2 arguments or more using double quotes was not working because of the regex


eSkiSo:
+1 on this, tried on master version and cant do for example:
 status:open tag:"DM" category:"Cat1" 

This will not show anything, if i remove either the tag: or the category: it will show but not with both, with this push it works correctly, shows task that has both the category and the tag.

PR #2808, opened 2016-11-10
closing 'X' for upload dialog

Busfreak:
The dialog 'Attach a document' in Template/task_file/create hides the
action buttons when a file is uploaded. You have to press ESC to close
the dialog. With this PR there is a closing 'X' in the right upper
corner.


fguillot:
Thank you for your contribution but I took another approach to implement this feature. I added the close button for all inline popups.

Busfreak:
Perfect!
Thanks!

PR #2897, opened 2016-12-20
array_combine error on task creation

Busfreak:
PHP 5.3 throws error
Warning: array_combine(): Both parameters should have at least 1 element in /var/www/ccc/kanboard/app/Core/ExternalTask/ExternalTaskManager.php on line 51

This PR fixes this :smile: 

Have a nice day!
array_combine

### DIFF
--- a/app/Core/ExternalTask/ExternalTaskManager.php
+++ b/app/Core/ExternalTask/ExternalTaskManager.php
@@ -48,6 +48,8 @@ class ExternalTaskManager
     public function getProvidersList()
     {
         $providers = array_keys($this->providers);
-        return array_combine($providers, $providers);
+        if (count($providers)) {
+            return array_combine($providers, $providers);
+        }
     }
 }


app/Core/ExternalTask/ExternalTaskManager.php
```diff
@@ -48,6 +48,8 @@ public function getProvider($name)
     public function getProvidersList()
     {
         $providers = array_keys($this->providers);
-        return array_combine($providers, $providers);
+        if (count($providers)) {
+            return array_combine($providers, $providers);
+        }
     }
```
fguillot: Thank you for reporting this bug but you forgot a return statement here, this function must always return an array. I fixed that in master, and now a test case ensure that no regression on this code will happen in the future.

Busfreak: Hi Frédéric,
you are so right, sorry, i've overlooked!
Thank you and have a happy new year!

PR #2938, opened 2017-01-12
Fix Markdown Syntax error

levlaz:
There was a syntax error where the link was broken to the MySQL
instructions.


fguillot:
Thanks!

Issue #2934, opened 2017-01-10
Editing Comment Does Not Update Timestamp


asftAndrew:
When editing a comment, the timestamp does not update to the date and time that the edit occurred. This makes it appears that the comment was made months ago when it was just recently updated.

I would expect that the date and time on the comment would be updated, however perhaps this could be an option set in the configuration file as to the behavior when editing comments.

levlaz:
Took a look at the schema, this would involve a DB change because the comments table does not track updated_date, only the `date_creation`. I think it would be useful to show both the created date and the updated date on the comment itself.

PR #2950, opened 2017-01-16
Comment update improvements

levlaz:
This fixes https://github.com/kanboard/kanboard/issues/2934

* Added date_modification to comments table
* Sort by date_modification instead of date_created 
* Show created and update dates in the UI 
* Added unit and integration tests (confirmed passing locally)

### DIFF
--- a/app/Schema/Postgres.php
+++ b/app/Schema/Postgres.php
@@ -11,6 +11,9 @@ const VERSION = 98;
 function version_98(PDO $pdo)
 {
     $pdo->exec('ALTER TABLE "comments" ADD COLUMN date_modification INT');
+    $pdo->exec('UPDATE "comments"
+        SET date_modificaiton = date_creation
+        WHERE date_modification IS NULL');
 }
 
 function version_97(PDO $pdo)
--- a/app/Schema/Postgres.php
+++ b/app/Schema/Postgres.php
@@ -6,7 +6,12 @@ use PDO;
 use Kanboard\Core\Security\Token;
 use Kanboard\Core\Security\Role;
 
-const VERSION = 97;
+const VERSION = 98;
+
+function version_98(PDO $pdo)
+{
+    $pdo->exec('ALTER TABLE "comments" ADD COLUMN date_modification INT')
+}
 
 function version_97(PDO $pdo)
 {
--- a/app/Schema/Postgres.php
+++ b/app/Schema/Postgres.php
@@ -10,7 +10,7 @@ const VERSION = 98;
 
 function version_98(PDO $pdo)
 {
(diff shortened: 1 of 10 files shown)


tests/units.postgres.xml
```diff
@@ -10,4 +10,4 @@
         <const name="DB_PASSWORD" value="" />
         <const name="DB_NAME" value="kanboard_unit_test" />
     </php>
-</phpunit>
\ No newline at end of file
+</phpunit>
```
levlaz: Bah! Sorry I can't seem to get this file back to the original state. :/

app/Template/comment/show.php
```diff
@@ -7,7 +7,9 @@
             <strong class="comment-username"><?= $this->text->e($comment['name'] ?: $comment['username']) ?></strong>
         <?php endif ?>
 
-        <small class="comment-date"><?= $this->dt->datetime($comment['date_creation']) ?></small>
+        <small class="comment-date">Created At: <?= $this->dt->datetime($comment['date_creation']) ?></small>
+        <small class="comment-date">Updated At: <?= $this->dt->datetime($comment['date_modification']) ?></small>
```
fguillot: The strings "Created At:" and "Updated At:" should be translatable.

app/Schema/Postgres.php
```diff
@@ -6,7 +6,12 @@
 use Kanboard\Core\Security\Token;
 use Kanboard\Core\Security\Role;
 
-const VERSION = 97;
+const VERSION = 98;
+
+function version_98(PDO $pdo)
+{
+    $pdo->exec('ALTER TABLE "comments" ADD COLUMN date_modification INT');
```
fguillot: You should use a BIGINT for Postgres as well, in this way we are ready for https://en.wikipedia.org/wiki/Year_2038_problem

app/Template/comment/show.php
levlaz: Thanks! Fixed.

app/Schema/Postgres.php
levlaz: Updated the migration file to do this.

app/Schema/Postgres.php
```diff
@@ -6,7 +6,15 @@
...
 use Kanboard\Core\Security\Role;
 
-const VERSION = 97;
+const VERSION = 98;
+
+function version_98(PDO $pdo)
+{
+    $pdo->exec('ALTER TABLE "comments" ADD COLUMN date_modification BIGINT');
+    $pdo->exec('UPDATE "comments"
+        SET date_modificaiton = date_creation
```
fguillot: Be careful, there is a typo here that prevent the migration to work :)

PR #2966, opened 2017-01-21
Disable auto-assignment in task dialogs

haggl:
In most cases you don't want to automatically self-assign newly created tasks.
Having to manually change the assignee field every time you want to create a new
task is cumbersome. Additionally, the "assign me" button is kind of useless when
the creator is pre-selected anyway. So the default should be "unspecified".


haggl:
Sorry - just realized that the patch also affected editing tasks after creation. Fixed.

fguillot:
This change have been already implemented in pull-request #2431.

Note: templates should not contains any logic.

PR #3041, opened 2017-02-15
updated german translation

Busfreak:


fguillot:
The unit tests doesn't pass because this translation is empty: `'The email project must be unique across all projects' => '',`

Busfreak:
UUh, sorry.
PR is updated to pass tests.

I wanted to translate, but I had an issue with this.
_'The **email project** must be unique across all projects'_
is the error message to
_'The **project email** is optional and could be used by several plugins.'_
so it should be:
_'The **project email** must be unique across all projects'_

What do you think?

fguillot:
Yep, I modified the string. Thanks!

Issue #2975, opened 2017-01-26
Feature request: Total number of score in this column across all swimlanes (like with tasks)


devBartek:
### Actual behaviour

In board view on top we have total number of tasks in column across all swimlanes. Label: "Total number of tasks in this column across all swimlanes"

### Expected behaviour

In board view on top we have sum of score in column across all swimlanes. Example: "Total number of score in this column across all swimlanes"

### Configuration
- Kanboard version: 1.0.37
- Database type and version: sqlite 3.8.10.2
- PHP version: 5.6.19
- OS: Windows NT 6.1
- Browser: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:51.0) Gecko/20100101 Firefox/51.0

fguillot:
This is now implemented by PR #3068

fguillot:
This is now available in Kanboard 1.0.40

PR #3068, opened 2017-02-22
Show total score across all swimlanes #2975

kvj:
Show total score, if available. 
Rendered to the left side of column/swimlane score, in brackets **`(12) 7`**


fguillot:
Thanks!

Issue #3052, opened 2017-02-17
Reintroduce word search in board selector from Kanboard 1.0.35


donpatzone:
### Actual behaviour

Trying to search for a word, which is not at the beginning in a multi-word project no result is returned
E.g. when searching for test in *This is a **test** project* will not return any results.

### Expected behaviour

Until KB 1.0.35 it was possible to search any word of a project in the board selector, granted you start with the first letter (Typing **est** instead of **test** will also not return anything)

### Steps to reproduce

Please supply step-by-step instructions so that anyone with the same environment can reproduce the issue. Whenever you offer a description, imagine
yourself in the shoes of someone who has never encountered this problem before and would like to know exactly how to observe it under a given set of conditions.
Here is an example:

1. Log in
2. Go to the board selector
3. Search for any word of a project, which is not at the beginning.

### Configuration

Copy and paste the configuration section from the Kanboard settings page or supply values for these fields:

- Kanboard version: 1.0.39
- Database type and version: SQLite
- PHP version: 5.6.29
- OS: Win2k8r2
- Browser: Firefox / Chrome (latest)

kvj:
That's the current behavior of `select-dropdown-autocomplete` component, regression happened when old component has been replaced with the new one. Can be easily fixed

fguillot:
Fixed by PR #3071

fguillot:
Available now in Kanboard 1.0.40

PR #3071, opened 2017-02-23
Reintroduce word search in board selector #3052

kvj:
Filter items of autocomplete dropdown by string inclusion, not just by prefix


fguillot:
Thanks !

PR #3075, opened 2017-02-25
Added hooks before and after the column header row

gusthoff:
Needed for ColumnGroup plug-in.


fguillot:
Can you update the documentation at the same time?

PR #3081, opened 2017-02-27
Almost done about spanish translation

EstebanMonge:
98% translated!!!


fguillot:
Thank you for your contribution, I synched the file with the latest version of Kanboard (apparently you send a file from an old version of Kanboard).

Here the updated version: https://github.com/kanboard/kanboard/blob/master/app/Locale/es_ES/translations.php

EstebanMonge:
Sorry, I made a fork with github, I will work on that.

PR #3110, opened 2017-03-10
Tasks expired do not update themself

TheCloud:
The TaskAssignColorOnDueDate  "touches" the task every time is expired so basically it looks like a change.
This prevents the automatic action that would remove tasks after X days of inactivity to be triggered.

### DIFF
--- a/app/Action/TaskAssignColorOnDueDate.php
+++ b/app/Action/TaskAssignColorOnDueDate.php
@@ -2,6 +2,7 @@
 
 namespace Kanboard\Action;
 
+use Kanboard\Action\Base;
 use Kanboard\Model\TaskModel;
 
 
@@ -48,6 +49,7 @@ class TaskAssignColorOnDueDate extends Base
         return array(
             'color_id' => t('Color'),
         );
+
     }
 
     /**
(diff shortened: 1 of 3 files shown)


TheCloud:
Fixed indent, sorry
About unit tests, it's about creating unit tests which tests compatibilities between different combinations of actions which is pretty complex. Infact we have noticed that setting colors on a expired tasks would changed the "modified" date so basically another action would not work as expected.

app/Action/TaskAssignColorOnDueDate.php
```diff
@@ -72,16 +74,15 @@ public function getEventRequiredParameters()
...
-                $values = array(
-                    'id'       => $task['id'],
-                    'color_id' => $this->getParam('color_id'),
-                );
-                $results[] = $this->taskModificationModel->update($values, false);
+            foreach ($data['tasks'] as $task) {
+                if (($task['date_due'] <= time() AND $task['date_due']>0) and $task['color_id']!=$this->getParam('color_id')) {
+	        $values = array(
+        	    'id' => $task['id'],
+		    'color_id' => $this->getParam('color_id'),
```
fguillot: This code doesn't seems to be properly indented.

app/Template/board/task_private.php
```diff
@@ -28,7 +28,7 @@
                     <?= $this->text->e($this->user->getInitials($task['assignee_name'] ?: $task['assignee_username'])) ?>
                 </span> -
             <?php endif ?>
-            <?= $this->text->e($task['title']) ?>
+            <span title="<?= $this->text->e($task['title']) ?>"><?= $this->text->e($task['title']) ?></span>
```
fguillot: This change doesn't seems to be related to this pull-request?

TheCloud: Yes they are two different things, don't know why they got merged into one.
I'm sorry I'm learning git :-)

fguillot:
Your changes have been integrated manually into master with this commit: https://github.com/kanboard/kanboard/commit/630f4ee780f3a9240ae9296e73f7518f21ae23ae

PR #3133, opened 2017-03-17
Small UI tweaks to improve tablet usage

k00ni:
This pull contains 2 minor changes to improve UI for mobile usage on tablets.

## **#board** 

Table with id=board uses the width given by the mother element. With that change (table-layout: fixed => inherit) the whole board uses only the "visible" width, avoiding scroll bars.

screenshot from 2017-03-17 14 39 10

## .sidebar-content

Class .sidebar-content has a different width, depending on the **min-width** value. With that change, the sidebar-content, such as task information, is place right of the sidebar, instead of dropping to the bottom.

screenshot from 2017-03-17 14 38 58

## General

Made to work with tablets, but using a smaller device width should be supported.

### DIFF
--- a/assets/sass/_board.sass
+++ b/assets/sass/_board.sass
@@ -11,7 +11,7 @@
   overflow-x: auto
 
 #board
-  table-layout: fixed
+  table-layout: inherit
   margin-bottom: 0
   th.board-column-header
     width: 240px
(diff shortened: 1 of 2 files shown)


fguillot:
I tried your PR on a iPad but your changes breaks the compact/horizontal mode of the board.

fguillot:
Closing stale PR.

k00ni:
Hi @fguillot: Sorry for late response. Currently i don't have the time to refine this PR, so no problem. For further proposals, are there any guidelines?

PR #3164, opened 2017-03-29
update to sqlite schema - fixing existing table

palica:
table project_activities contains a typo in schema, this corrects the typo on existing installations


fguillot:
Thanks!

PR #3183, opened 2017-04-07
Added mysql2 driver compatibility

boina-n:
Change the switch to make the application compatible with mysql2 driver


fguillot:
What is mysql2?

boina-n:
Mysql2 is a new project aimed for very fast mysql processing: https://github.com/brianmario/mysql2 .
I actually I tried to deploy Kanboard on the Pivotal Cloud Foundry PaaS then I found out that I wasn't using the right driver according to the switch in the code. actually my DB_DRIVER was mysql2. Also I see this project is compatible with Heroku, in order to make it compatible with pcfdev I had to adapt the code to mysql2.

PR #3199, opened 2017-04-16
update german translation

OHDMax:
I removed some inconsistencies like using the word "User" instead of "Benutzer" in the german translation


OHDMax:
I found another spelling error, which is not part of this PR but rather annoying since the text is displayed every time a task is created. It's in [line 1209](https://github.com/kanboard/kanboard/blob/master/app/Locale/de_DE/translations.php#L1209). Where it says "iHre" it should be "Ihre". Shall I create another PR for this typo?

fguillot:
I would prefer someone who speaks German to fix typos :)

fguillot:
So yes, you could send another PR.

OHDMax:
I do, it's my mother tongue ;)

PR #3213, opened 2017-04-24
Return HTTP 403 when access is forbidden

dopsi:
This would allow servers to handle wrong access (for example redirect to a custom error page).

### DIFF
--- a/app/Controller/AppController.php
+++ b/app/Controller/AppController.php
@@ -27,7 +27,7 @@ class AppController extends Base
             $this->response->html($this->helper->layout->app('app/forbidden', array(
                 'title' => t('Access Forbidden'),
                 'no_layout' => $withoutLayout,
-            )));
+            )), 403);
         }
     }
 


app/Controller/AppController.php
```diff
@@ -28,6 +28,7 @@ public function accessForbidden($withoutLayout = false, $message = '')
                 'title' => t('Access Forbidden'),
                 'no_layout' => $withoutLayout,
             )));
+            $this->response->withStatusCode(403);
```
fguillot: The status code should be set as second argument of the method html(), see https://github.com/kanboard/kanboard/blob/master/app/Core/Http/Response.php#L267

dopsi:
I have change that line.

dopsi:
I finally found out why the authentification was not working.

My Kanboard installation is located in a subdirectory of the domain (`example.com/kanboard/`) and the whole domain require HTTP Basic authentification.

If the `http_referer` was loaded from cache (in Chrome, for example when setting the page as homepage), the server does not set the `REMOTE_USER` variable because it has no authentification. Since nginx uses CGI to get the page and the CGI program does returns `HTTP 200` even though the script failed because of lacking authentification, the user cannot access the page but gets an *Access denied* message and the nginx reverse proxy cannot catch the error to request authentification.

PR #3249, opened 2017-05-09
Change Due Date to Due Timestamp fixes #919

### LINKED ISSUE #919: Change Due Date to Due Timestamp
For some tasks like meetings i would like to define a due timestamp and not only a due date. Can you change it please or add an additional due time field?
TIA && have a nice day

Basti

kiselli:
Added Time part to Task:date_due field in edit, view and export / import

### DIFF
--- a/app/Template/board/task_footer.php
+++ b/app/Template/board/task_footer.php
@@ -65,7 +65,7 @@
                 <?php endif ?>
                 ">
                 <i class="fa fa-calendar"></i>
-                <?= $this->dt->date($task['date_due']) ?> <?= $this->dt->time($task['date_due']) ?>
+                <?= $this->dt->datetime($task['date_due']) ?>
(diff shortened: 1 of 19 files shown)


app/Template/board/task_footer.php
```diff
@@ -65,7 +65,7 @@
                 <?php endif ?>
                 ">
                 <i class="fa fa-calendar"></i>
-                <?= $this->dt->date($task['date_due']) ?>
+                <?= $this->dt->date($task['date_due']) ?> <?= $this->dt->time($task['date_due']) ?>
```
fguillot: Instead of using the `time()` function, can you use directly the helper `$this->dt->datetime($task['date_due'])`?

kiselli:
Good idea, I changed the templates according to your suggestion.

PR #3261, opened 2017-05-13
Define a default filter phrase in config.php instead of the hard-coded 'status:open'

spst:
Instead of using the hard-coded default filter "status:open", I propose to define the default filter phrase in the config.php file. This addresses e.g. issue 1295

In addition it opens the possibility for a different approach to filter out the tasks not yet visible with the TaskBoardDate plugin.

Please note: I am not a developer. So there is a good chance that I did something wrong or in bad style. I will be happy to be corrected!

### DIFF
--- a/config.default.php
+++ b/config.default.php
@@ -237,3 +237,10 @@ define('HTTP_VERIFY_SSL_CERTIFICATE', true);
 
 // TOTP (2FA) issuer name
 define('TOTP_ISSUER', 'Kanboard');
+
+// default filter expression
+define('DEFAULT_FILTER', 'status:open');
+
+// an expression that will be added to the predefined filters
+// define('ADDITIONAL_FILTER', 'board_date:<=today');  // use this when plugin TaskBoardDate is installed
+define('ADDITIONAL_FILTER', '');  // use this otherwise
--- a/app/Core/User/UserSession.php
+++ b/app/Core/User/UserSession.php
@@ -176,7 +176,7 @@ class UserSession extends Base
      */
     public function getFilters($project_id)
     {
-        return ! empty($this->sessionStorage->filters[$project_id]) ? $this->sessionStorage->filters[$project_id] : 'status:open';
+        return ! empty($this->sessionStorage->filters[$project_id]) ? $this->sessionStorage->filters[$project_id] : DEFAULT_FILTER." ".ADDITIONAL_FILTER;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


spst:
Oh my...
this is not going so smooth.

Frédréric, could you indicate whether there is a chance, this will be merged? No definite decision, just an idea or rough guess.

If you think what I propose makes sense in a conceptional way, I will try to get it right. On the other hand, if your judgement is that it's not quite the right thing, I will leave it alone and close the pull request.

Best regards,
Stefan

... and sorry for stealing your time with my unpassed tests!

config.default.php
```diff
@@ -237,3 +237,10 @@
 
 // TOTP (2FA) issuer name
 define('TOTP_ISSUER', 'Kanboard');
+
+// default filter expression
+define('DEFAULT_FILTER', 'status:open');
+
+// an expression that will be added to the predefined filters
+// define('ADDITIONAL_FILTER', 'board_date:<=today');  // use this when plugin TaskBoardDate is installed
+define('ADDITIONAL_FILTER', '');  // use this otherwise
```
rautamiekka: Surely there's a way to detect TaskBoardDate's presence through PHP and just use ternary conditional here _like_
```php
// an expression that will be added to the predefined filters
// checks for the presence of TaskBoardDate plugin, adds a query, otherwise nothing.
define('ADDITIONAL_FILTER', isPresentTBD ? 'board_date:<=today' : '');
```
instead of relying on the System Admin ? If the SA **wants** to change anything on this line, he **will**, and the additional comment should tell him how to interpret that TC.

spst: In my view, the best option would be to leave this to the plugin, but I am not PHP-savvy enough to solve that...

including a mechanism that detects a plugin into kanboard's config is not really clean, in my opinion. It would ruin the separation between kanboard itself and plugins. On the other hand it would be a lot more convenient...

rautamiekka: ^ A plugin can't change Kanboard's config file, right ? In which case it'd have to be changed in the plugin's config.

fguillot: Kanboard should not be coupled and depends of any plugin.

Another way to do it is to offer the possibility to configure those filters via the user interface instead of using PHP constants. In this way, people do not have to manually change a config file on the server.

In the templates, you could use the helper `$app->config('my_config_param', 'some default value')`. Add new form fields in this template: https://github.com/kanboard/kanboard/blob/master/app/Template/config/project.php

rautamiekka: ^ Indeed. In some cases changing config files ain't even available to Kanboard Admins.

spst: Sorry for being unavailable for so long. I had (and actually still have) to focus my attention elsewhere...

I agree that core kanboard and plugins should not be mixed. So I guess, it's best to keep this idea as simple as possible, basically what the title says: "Define a default filter phrase in config.php", and let's ignore the board_date stuff for the moment.

I hope I will have more time in the near future to make some progress here...

spst:
Finally, I guess, I have to face it: I don't find enough time to do this properly. I would have loved to contribute, but I have to admit to myself, that it doesn't work out.

So I'll be closing the pull request. If someone wishes to pick up where I am leaving off, do feel free to do so. 

Best regards,
Stefan

PR #3298, opened 2017-06-10
add shadow to cards, different yellow

geromueller:
Didn't like to yellow and the border of the cards. I think it look more modern now. Cheers


rautamiekka:
Won't that just add more CPU usage ? And not everyone likes shadows.

PR #3305, opened 2017-06-17
Add default background white for the body tag

stephane-:
Hello,
On some old browsers the background appears black. The site is therefore not very usable. The fix does not change anything in modern browsers but fix the background color on some old browsers.
I do not submit my compiled css because I do not have the same configuration, and my compiler create lot of change on compiled file.

### DIFF
--- a/assets/sass/_reset.sass
+++ b/assets/sass/_reset.sass
@@ -22,7 +22,7 @@ body
   color: color('primary')
   font-family: $text-font
   text-rendering: optimizeLegibility
-  background-color: color('white') 
+  background-color: color('default') 
 
 small
   font-size: size('small')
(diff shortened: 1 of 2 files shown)


fguillot:
Which old browsers are you talking about?

stephane-:
Hello,
You can test it with connected TV (No smart) and more easily with steam browser (overlay or big picture).

Example : 
blackkanboard

assets/sass/_reset.sass
```diff
@@ -22,6 +22,7 @@ body
   color: color('primary')
   font-family: $text-font
   text-rendering: optimizeLegibility
+  background-color: color('white')
```
fguillot: It should be `color('default')` according to the change you made in `_variables.sass`?

stephane-:
Yes, sorry for that. I change the var name when I clean the code

fguillot:
Your commits have been merged into master.

PR #3315, opened 2017-06-26
Adding functionality for getting more info from reverse proxy auth

jeremyrogers:
As it was there was no way to get profile info from reverse proxy auth. I've added config elements for what the apache server should call the full name and the email. Then I've added Http/Request functions to get those values out. Then I've updated the ReverseProxyUserProvider class to take these new values in and return them.


fguillot:
Units tests don't pass.

fguillot:
Unfortunately, I cannot merge if the unit tests don't pass or if you introduce breaking changes.

PR #3343, opened 2017-07-11
Few fixes and improvements

TheCloud:
76742e8
Bug #2681 when filter applied, now d&d drops elements where expected without magic changes. Still need more work to do to get it 100% perfect but community seems happy with it now

60f4181
When automatic action is applied on expired items, they get touched/updated so they never expire. This change make let them expire

49ac828
when board is in collapsed view mode, titles are not fully readable, so MouseOver show the title

5050803
Never let a task be created without it's creator (maybe obsolete?)


TheCloud:
I would love to, but don't know how to separate PRs

psy-q:
Just separate them into different branches (don't do everything on master). It's good practice anyway because if you break the build you can tell which feature introduced the problem.

app/Action/TaskAssignColorOnDueDate.php
```diff
@@ -81,7 +83,6 @@ public function doAction(array $data)
                 );
                 $results[] = $this->taskModificationModel->update($values, false);
             }
-        }
```
fguillot: This curly brace is necessary.

PR #3345, opened 2017-07-11
Add "Create another link" checkbox for internal link creation

oliviermaridat:
Add "Create another link" checkbox for internal link creation as in sub-task creation.

This is really useful when you have one task which is "blocking" / "is a milestone of" / "relates to" several other tasks.

I check the implementation for sub-task and tried to mimic as far as possible. This is why I also introduced a "controller:tasklink:form:default" hook (which is not the main goal of this PR).

Thanks!

### DIFF
--- a/app/Controller/TaskInternalLinkController.php
+++ b/app/Controller/TaskInternalLinkController.php
@@ -44,6 +44,13 @@ class TaskInternalLinkController extends BaseController
     {
         $task = $this->getTask();
 
+        if (empty($values)) {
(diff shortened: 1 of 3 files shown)


oliviermaridat:
Build is only not passing for PHP 5.3 DB=mysql... The job was more than 90% but then "The job exceeded the maximum time limit for jobs, and has been terminated.".

Therefore, this is not a build failed but a build timeout issue.

PR #3361, opened 2017-07-20
Project User Roles are "Restrictions"

drewlloyd:
Reversed the if statement to correct the restriction logic


drewlloyd:
test fail, more testing needed

PR #3416, opened 2017-08-28
Fixing LDAP authetication problem

jmbatistav:
Fixing LDAP authentication problem, because it used the login LDAP user password with the config.php file LDAP user.

### DIFF
--- a/app/Auth/LdapAuth.php
+++ b/app/Auth/LdapAuth.php
@@ -153,7 +153,7 @@ class LdapAuth extends Base implements PasswordAuthenticationProviderInterface
             case 'proxy':
                 return LDAP_PASSWORD;
             case 'user':
-                return $this->password;
+                return sprintf(LDAP_PASSWORD, $this->password);
             default:
                 return null;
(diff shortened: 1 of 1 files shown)


fguillot:
What problem are you trying to solve?

jmbatistav:
When you configure the LDAP authentication in the config.php file, you use the LDAP_PASSWORD and LDAP_USERNAME vars for 'user' or 'proxy' LDAP bind type. So, when you try to authenticate from the login form, the system use the username that came from the form with the password stored in LDAP_PASSWORD var and try to authenticate, but this behavior is wrong. Sorry for my bad english.

jmbatistav:
Sorry, my bad, it takes the user stored in LDAP_USERNAME var and the password that came from the login form and try to authenticate the first time before searching in LDAP directory.

fguillot:
I still don't know why you need to sprintf the password... For me, your are introducing a bug/breaking change.

fguillot:
Closing stale PR.

PR #3439, opened 2017-09-05
translation optimize

zhl006006:
translation optimize


fguillot:
What are you trying to optimize?

You could send a separate pull-request just for the translations.

fguillot:
The translation has been updated manually.

PR #3467, opened 2017-09-25
Fix upload

kent1D:
Allow to add and remove files fo/from the queue before upload


kent1D:
Should Fix#3466

fguillot:
Why did you duplicate this file `assets/js/components/file-upload.js` to `assets/js/src/file-upload.js`?

kent1D:
oh sorry.

A problem of copy pasting and attention when I've sent the code.

I will change it asap.

PR #3485, opened 2017-10-10
added configuration volume to Dockerfile

Lerentis:
Hello there,

while working with your Docker Container i noticed, that the configuration is not persistent and should be excluded to another volume.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,15 @@
 FROM fguillot/alpine-nginx-php7
 
 COPY . /var/www/app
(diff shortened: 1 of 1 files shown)


fguillot:
You can already store your own config file in the data volume: 

- https://github.com/kanboard/kanboard/blob/master/doc/en_US/docker.markdown#config-files
- https://github.com/kanboard/kanboard/blob/master/app/common.php#L24-L28

Lerentis:
@fguillot uh damn didn't read that. sorry to have wasted your time

PR #3501, opened 2017-10-19
Improved display of the header bar

yannicka:
Before:
screenshot-2017-10-19 tableau de bord pour admin 2

After:
screenshot-2017-10-19 tableau de bord pour admin

- Elements on the header bar are vertically centered
- A (light) gray has been added in the background
- The margin is no longer in the `body`
- Displays bullets from `<ul>` in tooltips


fguillot:
I think there is a regression with your changes. If you reduce the size of the window to simulate a mobile device, the user menu dropdown is under the board switcher instead of being above.

Current behavior:

image

Your changes:

image

yannicka:
Fixed:
capture d ecran - 2017-10-20 a 22 45 36

fguillot:
Thanks

PR #3503, opened 2017-10-21
Remove most of the border-radius

yannicka:
This remove most of the border-radius. I know it will be controversial, but I try anyway. I am open to discussion.

Example:
capture d ecran - 2017-10-21 a 21 03 14


fguillot:
What is the problem with border-radius?

yannicka:
There's no real problem, I just find them superfluous, I think they make the interface less beautiful.

fguillot:
Perhaps, using a custom CSS stylesheet that remove the border-radius will avoid a controversial discussion?

yannicka:
Yes, you are right. I'll do that.

PR #3515, opened 2017-10-27
Ignore webhook notification if X-Skip-Webhooks HTTP header is set

dkrmr:


bmalynovytch:
This feature is used when webhooks are used to sync Kanboard with another tool, which also emits webhook notifications.
With this header set by the first request, we can block cycle propagation of the events.

app/Notification/WebhookNotification.php
```diff
@@ -35,6 +35,10 @@ public function notifyUser(array $user, $event_name, array $event_data)
      */
     public function notifyProject(array $project, $event_name, array $event_data)
     {
+        if(!empty($_SERVER['HTTP_X_SKIP_WEBHOOKS']) && $_SERVER['HTTP_X_SKIP_WEBHOOKS']) {
```
fguillot: 1. You should not use global variables. There is a class `Kanboard\Core\Http\Request` for that.
2. You should add tests
3. You should update the documentation if you add new features (folder `doc`)

fguillot:
Closing stale PR after 30 days.

PR #3545, opened 2017-11-12
ignore vendor folder

igbanam:
For development purposes, this folder could be ignored. Since all
contents of this folder is generated from `composer.json`. Corollaries to
this are how `node_modules` and `bower_components` are ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ data/cache
 !docker/kanboard/config.php
 node_modules
 bower_components
+vendor


fguillot:
This folder has been included intentionally in the repo since this commit https://github.com/kanboard/kanboard/commit/9e2b2a32fd0e967ad3184e9a5d091a29953acb91

- This is mentioned in the ChangeLog
- That allow people to download a ready to use archive by using git-archive or the download link of Github.

zorn-v:
Maybe there is should be a script that generate "ready to use" archive ?
Now, to check changes for PR (consider run test) you MUST run `composer install` and then clear vendor dir of dev dependents (consder remove it and run `git reset --hard`).
It is not conveniently.

Issue #2444, opened 2016-07-06
Feature request: case insensitive usernames on login


Biniou180:
I open this ticket as a contributor of Framasoft projects, and particularly Framaboard which is based on Kanboard.

Currently (1.0.25 on framaboard), usernames are case sensitive when logging in.
It can be great, but we have a lot of support requests concerning users unable to connect because they forgot they had uppercase characters in their usernames.
It could be great to ease the login process for our users :)
### Steps to reproduce
1. Create a user including at least one uppercase character
2. Log out
3. Forgot that the username has an uppercase character :)
4. Log in with lowercase username
### Suggested solutions
- To indicate at least that the username is case sensitive on the login page
- Best: Forbid uppercase in usernames (rather common for online services)
- Much better: Allow the user to login with his e-mail address too
- The best: Allow the user to reset his password by giving his e-mail address (the reply e-mail will contain the username)

Thanks a lot!

Biniou180:
Hello, 
I just put a comment here to ask if this issue is planned in a future version by chance?
I have just discussed with one of our users about this ticket, particularly the ability to reset the password by giving e-mail, as final users often not remember their username but definitely remember their e-mail :)

Thanks again!

bugwatcher:
This issue has been closed automatically by a bot for inactivity. If you still need this feature, you should consider contributing to this open-source project. For more information, read the [FAQ](https://docs.kanboard.org/en/1.2.5/developer_guide/faq.html#why-do-you-close-inactive-issues-automatically).

PR #3554, opened 2017-11-15
Sqlite username case insensitive

zorn-v:
One time I accidently created two accounts "zorn" and "Zorn" (with same password) using sqlite DB. And succesfully forgot about it. You can imagine my frustration when I logged in with "wrong" account and did not see any tasks related to me 😄 

### Warning
After executing this migration on database with usernames differs only in case  there will be undefined behavior.

I think it should be in notes for release if this PR will merged.

### DIFF
--- a/app/Schema/Sqlite.php
+++ b/app/Schema/Sqlite.php
@@ -8,7 +8,41 @@ use Kanboard\Core\Security\Token;
 use Kanboard\Core\Security\Role;
 use PDO;
 
-const VERSION = 116;
+const VERSION = 117;
+
+/** Makes username case insensitive */
+function version_117(PDO $pdo)
+{
+    $pdo->exec("CREATE TABLE _users_new (
+        id INTEGER PRIMARY KEY,
+        username TEXT NOT NULL COLLATE NOCASE UNIQUE,
+        password TEXT,
+        is_admin INTEGER DEFAULT 0,
+        is_ldap_user INTEGER DEFAULT 0,
+        name TEXT, email TEXT,
+        google_id TEXT,
+        github_id TEXT,
+        notifications_enabled INTEGER DEFAULT '0',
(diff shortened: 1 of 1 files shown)


fguillot:
Instead of changing the schema for Sqlite, a more generic solution for all supported databases should be implemented.

It could be simple as adding an config parameter that lowercase the username if enabled. There is an open feature request for that #2444.

zorn-v:
I don't think so. In all databases (except sqlite) collation is CI usually by default.
More code, more potential bugs.

fguillot:
This option already exists for LDAP authentication. This is a simple `strtolower()`: https://github.com/kanboard/kanboard/blob/master/app/User/LdapUserProvider.php#L168

You could do something similar instead of taking the risk of changing the schema.

rautamiekka:
And I just checked that both MySQL and SQLite (no idea about Postgre) have the `BINARY` keyword in the `WHERE` clause which is used to force case-sensitivity for that part of the query, so you could have the db default to case-insensitive and check a setting of case-sensitivity in Kanboard's config, deciding whether to use `BINARY` keyword.

I've used that keyword in some cases in MySQL.

zorn-v:
@fguillot It is NOT option. It's some constant.
@rautamiekka I dont understand in which cases you should have usernames with case sensitive. Show me at least one service/system in which you can have two different users with usernames differs only in case. It's provides confusion. Why it should be a option ?
Are you shure that usernames check with BINARY on login ?

zorn-v:
@fguillot that line of code get username of already fetched user I think. You can not login anyway.

fguillot:
Your solution have to work with all supported database anyway.

zorn-v:
No, ok. I did it for my DB and NOW I dont care if anyone faced with this problem

PR #3633, opened 2017-12-22
Sanitize strings

urbalazs:
Hi @fguillot 
This PR removes the colons and spaces from the end of the strings. This reduces the amount of translatable strings, and avoid wording errors caused by missing space or colon from the translations.

### DIFF
--- a/app/Template/analytic/lead_cycle_time.php
+++ b/app/Template/analytic/lead_cycle_time.php
@@ -6,8 +6,8 @@
 
 <div class="panel">
     <ul>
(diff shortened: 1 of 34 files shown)


fguillot:
The colons must be part of the translated string because it can be different for some language. For example, in French we write `Auteur :` not `Auteur:` for the English string `Author:`. Quotes and or other punctuation marks are not necessary the same either.

urbalazs:
I didn't know this. Now I leave strings as they are.

PR #3645, opened 2017-12-29
Missing translation for 'sign up'

sfahrenholz:
This PR contains the text string "sign up" from the login screen. This values was not in the translation files. so i added them and give a translation for german.


fguillot:
I cannot find the string `Sign up`  in Kanboard core (only `Sign-up` which is already translated). Perhaps you are looking to translate the Registration plugin? (https://github.com/kanboard/plugin-registration/tree/master/Locale)

Issue #3026, opened 2017-02-10
Feature request: Add plugin hook after creating task returning ID of created task


RAZORzdenko:
I built small plugin to allow organizations to be added to projects and assigned to individual tasks. I already have it built into my local kanboard fork and I will perform pull request if there's interest in this. It would be nice to get it officially included without worrying about compatibility issues in future.

fguillot:
Well, send a pull-request and we will see...

PR #3660, opened 2018-01-09
Feature request #3026: Added plugin hook 'aftersave' after creating Task

RAZORzdenko:
This plugin hook returns ID of currently created task

### DIFF
--- a/app/Model/TaskCreationModel.php
+++ b/app/Model/TaskCreationModel.php
@@ -47,7 +47,7 @@ class TaskCreationModel extends Base
             ));
         }
 
-        $this->hook->reference('model:task:creation:aftersave', (int) $task_id);
+        $this->hook->reference('model:task:creation:aftersave', $task_id);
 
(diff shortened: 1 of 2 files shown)


fguillot:
If you add a new hook, can you update the docs as well?

RAZORzdenko:
Just updated it. Hopefully it's in line with contributing guidelines. Thank you for patience.

kent1D:
It’s after creating ? Or after each save of tasls? For the documentation?

kent1D:
It would be nice to have one after 1st creation and a different one after each save. Maybe with just a different parameter?

RAZORzdenko:
This hook is only available after creating task. If you need to get task or something else after updating a task, you can look at TaskModificationModel.php and add your hook there.

https://github.com/kanboard/kanboard/blob/daf39ee16aa15572b92f96bd7cc3705305a50b9d/app/Model/TaskModificationModel.php

PR #3731, opened 2018-02-22
[Enhancement] category-Tag customizing CSS

sfahrenholz:
With this PR it is possible to create specify styles for every tag.
How it works is documented in "individuel-category.markdown"


app/Template/board/task_footer.php
```diff
@@ -1,6 +1,6 @@
 <?php if (! empty($task['category_id'])): ?>
-<div class="task-board-category-container">
-    <span class="task-board-category">
+<div class="task-board-category-container task-board-category-container-color">
+    <span class="task-board-category tag-<?= $this->text->e($task['category_name']) ?>">
```
fguillot: Using `category-` instead of `tag-` might be more appropriate because Kanboard have tags and categories?

sfahrenholz: you are right, i changed the "tag-" value, also in the markdown informations. thanks for the information, i didnt notice this inconsistency!

PR #3739, opened 2018-02-27
make it possibe to configer reply-to, authorname, in send

kwibus:
Hi, I am working on a plug-in for kanboard to extend mail functionality. 
To make kanboard usable for us ([at computing](www.atcomputing.nl)) system for remote support.  Our plug-in would work better if you could optional  configure reply-to, author name, in send function.

this patch keeps the original behaviour of send function. but if  you can supply it with extra arguments, to make it more configurable.
  
We could do without this patch. But we hope you like it, Comments our welcome.

### DIFF
--- a/app/Core/Mail/Client.php
+++ b/app/Core/Mail/Client.php
@@ -44,7 +44,7 @@ class Client extends Base
      * @param  string  $html
      * @return Client
      */
-    public function send($recipientEmail, $recipientName, $subject, $html, $authorName=null, $authorEmail=null)
+    public function send($recipientEmail, $recipientName, $subject, $html, $authorName = null, $authorEmail = null)
     {
         if (! empty($recipientEmail)) {
             $this->queueManager->push(EmailJob::getInstance($this->container)->withParams(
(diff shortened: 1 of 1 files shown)


app/Core/Mail/Client.php
```diff
@@ -44,16 +44,16 @@ public function __construct(Container $container)
      * @param  string  $html
      * @return Client
      */
-    public function send($recipientEmail, $recipientName, $subject, $html)
+    public function send($recipientEmail, $recipientName, $subject, $html, $authorName=null, $authorEmail=null)
```
fguillot: Please, add a space around the operator `=`.

app/Core/Mail/Client.php
```diff
@@ -44,16 +44,16 @@ public function __construct(Container $container)
...
     {
         if (! empty($recipientEmail)) {
             $this->queueManager->push(EmailJob::getInstance($this->container)->withParams(
                 $recipientEmail,
                 $recipientName,
                 $subject,
                 $html,
-                $this->getAuthorName(),
-                $this->getAuthorEmail()
+                (is_null ($authorName) ? $this->getAuthorName() : $authorName),
```
fguillot: The first and last parentheses are not necessary. Can you change `is_null (` to `is_null(`?

kwibus:
Thank you for the feedback. Hope it's good now.

fguillot:
I was about to merge but I don't see your latest changes in the pull-request?

kwibus:
Sorry my fault, pushed the wrong branch, second try

PR #3746, opened 2018-03-02
Redirect to original URL after oauth login

### LINKED ISSUE #3684: Loosing linked URL after oauth login
This software focus on simplicity, the number of features is voluntary limited.

### Actual behaviour

When we click on a shared URL (mail, notification) and need to log in via oauth (gitlab in our case), we are redirected to the dashboard after login.

### Expected behaviour

To be redirected to the original URL

kent1D:
Should FIX #3684

### DIFF
--- a/app/Controller/OAuthController.php
+++ b/app/Controller/OAuthController.php
@@ -106,7 +106,7 @@ class OAuthController extends BaseController
     protected function authenticate($providerName)
     {
         if ($this->authenticationManager->oauthAuthentication($providerName)) {
-            $this->response->redirect($this->helper->url->to('DashboardController', 'show'));
+            $this->redirectAfterLogin();
         } else {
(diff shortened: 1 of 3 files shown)


app/Controller/OAuthController.php
```diff
@@ -127,4 +127,20 @@ protected function authenticationFailure($message)
             'title' => t('Login')
         )));
     }
+
+    /**
+     * Redirect the user after the authentication
+     *
+     * @access private
+     */
+    private function redirectAfterLogin()
```
fguillot: Please, do not duplicate code.

Move this method from `AuthController` to the `BaseController` and make it protected.

kent1D: done

kent1D:
Thank you ;)

Issue #3413, opened 2017-08-27
How to remove standard filter "status:open"?


minthemiddle:
### Actual behaviour

On all project views, the filter `status:open` is set automatically. I have to manually remove it to see my closed tasks.

### Expected behaviour

I would expect to set the default view filter somewhere instead having to manually remove a default one that does not suit my workflow. I prefer to see my closed tasks so I can visually get how much work I finished on a given day.

El-Gavy:
I agree with that, i have the same need.
Working as a team, i want my coworkers to be able to see finished tasks without having to explain them how to manually remove the filter each time they connect.

bugwatcher:
This issue has been closed automatically by a bot because it was tagged as **"question"**. Please, use the [forum](https://github.com/kanboard/forum/issues) for open discussions to avoid polluting the bug tracker. Thank you for your understanding.

Issue #3225, opened 2017-04-30
How do I Change Board/Project Default Filter


seancarson:
**Please, do not create duplicate issues**

### Actual behavior

When opening a new or existing board the filter defaults to Status:Open.  

### Expected behavior

In my kanboard instance I would like to change the default filter for all boards to Status:Open Status:Closed

### Steps to reproduce

1. Log into kanboard as administrator or regular users. 
2. Create or open and existing project.  
3. You will now see a text  area/filter with status:open

 I attempted to follow some of the example set in request https://github.com/kanboard/kanboard/issues/1295
but this does not apply to the version I am using.  

### Configuration

Copy and paste the configuration section from the Kanboard settings page or supply values for these fields:

- Kanboard version: 1.0.42
- Database type and version: postgres 9.3.16
- PHP version: 5.5.9-1ubuntu4.21
- OS:Linux 3.13.0-116-generic
- Browser: Chrome/57.0.2987.133

Thanks

SC

fguillot:
See https://github.com/kanboard/kanboard/blob/master/app/Core/User/UserSession.php#L179

Note: It's more appropriate to post your questions on the forum instead of the bug tracker

seancarson:
will do next time.  Thank You

On Apr 30, 2017, at 12:06 PM, Frédéric Guillot <notifications@github.com<mailto:notifications@github.com>> wrote:


See https://github.com/kanboard/kanboard/blob/master/app/Core/User/UserSession.php#L179

Note: It's more appropriate to post your questions on the forum instead of the bug tracker

lohierf:
Thank you, I was also looking to change this default filter to "status:all" instead of "status:open".

For a better user experience, I also changed the default filter of the "Reset filters" button to "status:all" : https://github.com/kanboard/kanboard/blob/master/app/Template/project_header/search.php#L11

Issue #2222, opened 2016-05-13
Ability to Set Default Filters per Board


seancarson:
Create option under settings that allows the board admin to set the default filter in a board.  

It is currently defaulted to status:open.  

A more advance option down the road may be the ability to lock the board to the default filter. Basically grey out the filter option for non project manager users.   

These features would allow Kanboard to be more easily used in various use cases such as helpdesk, or even bug reporting.  **creator:me**  for instance is a very good one and if it was defaulted and locked would make kanboard an excellent helpdesk system.  Of course after you created all the user accounts as local or remote users  lol.  

Thank you for writing Kanboard.

seancarson:
Also, if it makes more since to set it globally a good location to place this feature maybe Settings > Project settings.  this is where all the other global project settings exist.

Thanks

LordFPL:
+1 :)

erpnedir:
If the default filters can be defined **per board per user role**, and we have the option **to lock the filter** it would solve lots of my restriction and reporting purposes. 
+1 for this feature request

ruihildt:
Also needed here. :)

bugwatcher:
This issue has been closed automatically for inactivity after more than 1 year. If you still need this feature, you should consider contributing to this open-source project.

Issue #1323, opened 2015-10-05
Setting option: Default Filter query


starjack1:
The default Filter query in the Project view is: status:open

Maybe there could be a Setting option, to choose his own default query. (in my case an empty query, want to see the closed tasks by default.)

Simple Kanboard with excellent connectivity options. Love your work.

BlueTeck:
related #1295

bugwatcher:
This issue has been closed automatically for inactivity after 2 years. If you still need this feature, you should consider contributing to this open-source project.

PR #3781, opened 2018-03-21
Add default filter per user

AurelienC:
Each user can have a default filter rather than `status: open`.

This pull request is related to issues #1323, #2222, #3225, #3261 and #3413.


AurelienC:
I will check CI errors.

PR #3784, opened 2018-03-21
Add default filter per user

AurelienC:
Each user can have a default filter rather than `status: open`.

This pull request is related to issues #1323, #2222, #3225, #3261 and #3413.

### DIFF
--- a/app/Core/User/UserSession.php
+++ b/app/Core/User/UserSession.php
@@ -223,7 +223,8 @@ class UserSession extends Base
     public function getFilters($projectID)
     {
         if (! session_exists('filters:'.$projectID)) {
-            return 'status:open';
+            // return 'status:open';
+            return session_get('user')['filter'];
         }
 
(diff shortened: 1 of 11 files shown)


app/Schema/Mysql.php
```diff
@@ -8,7 +8,12 @@
 use Kanboard\Core\Security\Token;
 use Kanboard\Core\Security\Role;
 
-const VERSION = 127;
+const VERSION = 128;
+
+function version_188(PDO $pdo)
```
fguillot: 188?

app/Core/User/UserSession.php
```diff
@@ -223,7 +223,8 @@ public function isLogged()
     public function getFilters($projectID)
     {
         if (! session_exists('filters:'.$projectID)) {
-            return 'status:open';
+            // return 'status:open';
```
fguillot: There is no need to have commented code.

PR #3800, opened 2018-04-03
update doc

stuba:
Hi,

current documentation for API task method moveTaskPosition seems to give a Invalid params -error. Here is updated working sample.


fguillot:
I am in the process of moving the documentation to https://docs.kanboard.org/ so I reflected the change here: https://docs.kanboard.org/en/latest/api/task_procedures.html#movetaskposition

Thanks

PR #3823, opened 2018-04-18
Update pt-BR translation

filhocf:
Update pt-BR translation.


filhocf:
Is true. I will cancel this PR and I will do a new. Thanks by warning.

Issue #3857, opened 2018-05-11
Time formatting code in two additional templates may need updating


sudoman:
### Actual behaviour

In commit eafaa4b03770a73080520952112905eb8135a62f, hard-coded time format strings were updated, but it looks like these files were missed:

````
app/Template/notification/subtask_create.php
app/Template/notification/subtask_update.php
````

Grepping through the code didn't show me where those templates are in use, if at all.

### Expected behaviour

These templates should not have hard-coded time format strings.

### Configuration

- Kanboard version: master (eafaa4b03770a73080520952112905eb8135a62f)
- Database type and version: Sqlite3
- PHP version: 7
- OS: GNU/Linux
- Browser: Firefox

sudoman:
Correction, only one Template was missed:

````
app/Template/notification/subtask_create.php
````

I'm working on a pull request for it.

fguillot:
PR merged thanks

PR #3858, opened 2018-05-11
translate time string in another template

sudoman:

I wasn't to live test this change because I didn't find the precise location that this template is used. I tested the notifications for new subtasks, but the displayed text is different than the text in this template.


sudoman:
This pull request resolves #3857

Issue #3188, opened 2017-04-10
Feature request: Printing/exporting the board view


rolandkoller:
### Actual behaviour

When printing the board view, the output is text only, the board is not visually reproduced.

![screenshot printing](https://cloud.githubusercontent.com/assets/27192481/24857860/0d150bdc-1dea-11e7-98d2-caf4ab2054fd.png)

### Expected behaviour

It would be nice to have the posibility to export (PDF) or print a graphical representation of the board view.

![screenshot board view](https://cloud.githubusercontent.com/assets/27192481/24857861/0d15772a-1dea-11e7-91bb-b4b231a4877f.png)

### Steps to reproduce

Press Ctrl-P in e. g. IE, Firefox in board view

tzeumer:
I think printing does not make to much sense, but some co-workers want to print boards anyway, too. One gets a descent print layout if the style were not explicitly applied for media=screen only. An ugly hack is this

1. Edit {kanboard-folder}\app\Template\layout.php
2. Replace " <?= $this->asset->css('assets/css/app.min.css') ?>" with
```
<!-- Replacement code. Printing hack. Landscape, keep tasks together on page, show overflow -->
<style type="text/css"  media="print">
    @page {
        orientation:landscape;
        -webkit-transform: rotate(-90deg); 
        -moz-transform:rotate(-90deg);
        filter:progid:DXImageTransform.Microsoft.BasicImage(rotation=3);
    }
    #board-container { overflow-x: initial !important; }
    .board-task-list { min-height: 0 !important; }
    .task-board      { page-break-inside: avoid; }
</style>
<?= $this->asset->css('assets/css/app.min.css', true, 'screen, print') ?>
<!-- Original code; <?= $this->asset->css('assets/css/app.min.css') ?> -->
```

It's not perfect but works well for a reasonable number of columns. Maybe it helps someone.

rolandkoller:
@tzeumer: Your css workaround is quite good. Thank you!

rajo:
@tzeumer this works very well
@fguillot could this workaround become an official solution?

lastlink:
I might also hide the menu's in the css
```css
@media print {
  /*Hide menu */
  .menu-inline, .project-header, .menus-container { display: none; }
}
```

lastlink:
This is my final hiding menu's from printing. May still need work

```css
@media print {
@page {
        orientation:landscape;
        -webkit-transform: rotate(-90deg); 
[...]

PR #3875, opened 2018-05-21
add global print sass file

lastlink:
## Notes:
will need to regenerate app.css.min, decided not to call _print.sass b/c print in app.sass looked like a keyword.

-  [x] I read the [contributor guidelines](https://docs.kanboard.org/en/latest/developer_guide/contributing.html#i-want-to-contribute-to-the-code)

## Fixes:
* #3188

### DIFF
--- a/assets/sass/_global_print.sass
+++ b/assets/sass/_global_print.sass
@@ -1,0 +1,18 @@
+@import variables
+@import mixins
+	
+@media print {
+@page {
+        orientation:landscape;
+        -webkit-transform: rotate(-90deg); 
+        -moz-transform:rotate(-90deg);
+        filter:progid:DXImageTransform.Microsoft.BasicImage(rotation=3);
+    }
+    #board-container { overflow-x: initial !important; }
+    .board-task-list { min-height: 0 !important; }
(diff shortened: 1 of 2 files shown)


lastlink:
@fguillot what command are you using to build it? So I can test & fix it

fguillot:
I run `gulp css`, take a look to this page for the installation: https://docs.kanboard.org/en/latest/developer_guide/assets.html

lastlink:
Ok, I validated and fixed w/ the gulp css command thanks. Should build now.

fguillot:
The compilation works but the CSS doesn't seems to be taken into consideration when you open the print window in Chrome. Perhaps the Sass syntax for print media query is incorrect.

lastlink:
@fguillot Ok, I've fixed and **tested** that the media @print is working.
I updated the gulp css build & default to accommodate the difference for print.
I made the discovery in `layout.php` that the default for media app.min.css was 'screen'. So I had to separate the .min completely to use the media print.
        `<?= $this->asset->css('assets/css/print.min.css', true, 'print') ?>`

PR #3886, opened 2018-05-29
make border around recently modified tasks more visible

stinnux:
The small 2px frame around recently changed tasks is barely visible, so i changed the css to make it thicker and dashed.


fguillot:
In my opinion, a border of 5px is too big. You can use the custom CSS feature for that kind of things.

PR #3890, opened 2018-05-31
add daily background job to cron under docker

Kolossi:
:white_check_mark: I read the [contributor guidelines]

Under docker the daily background task doesn't seem to be set up.

If I run `./cli cronjob` manually all functionality kicks in fine.

I believe adding this file will fix it, but I'm not able to build and test a docker at present (I'm working on it...)

Another way to do this would have been to add the file to `/etc/periodic/daily`, but I believe it would then run as root and so might not be ideal.  Happy to change to this if that's preferred though.

Please review and if you agree merge it and I'll check the docker image on next build.

### DIFF
--- a/docker/etc/crontabs/nginx
+++ b/docker/etc/crontabs/nginx
@@ -1,0 +1,1 @@
+0 8 * * * cd /var/www/app && ./cli cronjob >/dev/null 2>&1


fguillot:
The cronjob should already work, see: https://github.com/kanboard/kanboard/blob/master/docker/var/spool/crontabs/nginx

Kolossi:
Hmm I hadn't seen that.

But it doesn't seem that `/var/spool` is hooked up to anything.

If I do `crontab -u nginx -e` then it gets placed in `/etc/crontabs/nginx`.

I also tried adding
`* * * * * date >> /tmp/crontest.etc` to `/etc/crontabs/nginx` (with crontab -e)
and
`* * * * * date >> /tmp/crontest.var` to `/var/spool/crontabs/nginx`

and yes only `/tmp/crontext.etc` got created.

So it would seem from this that my proposed change would fix it up and `/docker/var/spool` should be removed?

I don't _think_ it's the case but perhaps it depends on the underlying docker host?  I run it on the Docker package on a Synology DS415+ NAS.  It would kind of defeat the object of docker if that were the case though.

So I guess the question is - "is anyone running the current docker image and getting the daily background job running ok?"

Kolossi:
... the above all from within the container via
`docker exec -it kanboard bash`

Kolossi:
UPDATE - Last night I had
* `/var/spool/crontabs/nginx` with `1 0 * * * ...` schedule
* `/etc/crontabs/nginx` with `0 8 * * * ...` schedule

The job fired only at 08:00

PR #3925, opened 2018-06-21
Propose YunoHost badge for installing

anmol26s:
Hi
The YunoHost have working Kanboard package with  working install,remove,update,backup and restore scripts. YunoHost makes self -hosting easy for the end users. Please add the this badge so that people can find this solution.

[] I read the [contributor guidelines](https://docs.kanboard.org/en/latest/developer_guide/contributing.html#i-want-to-contribute-to-the-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,11 @@ Kanboard is project management software that focuses on the Kanban methodology.
     - [Installation instructions](https://docs.kanboard.org/en/latest/admin_guide/installation.html)
     - [Upgrade to a new version](https://docs.kanboard.org/en/latest/admin_guide/upgrade.html)
     - [Use Kanboard with Docker](https://docs.kanboard.org/en/latest/admin_guide/docker.html)
+ 
+ ## Run on YunoHost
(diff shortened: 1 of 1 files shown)


fguillot:
I would prefer having this in the documentation instead of the readme. Same as other installation options: https://docs.kanboard.org/en/1.2.5/admin_guide/index.html

anmol26s:
But for YunoHost there is no need to have a document. Its just click and install for the YunoHost users.
What would be written in the installation guide for the YunoHost page?
If you don't feel good about the badge then we can just put this line.
https://github.com/kanboard/kanboard/pull/3925/commits/393597d92f24d4b5a740400e79c1751ba375f400#diff-04c6e90faac2675aa89e2176d2eec7d8R20

fguillot:
You could do something similar as the Heroku documentation page, there is an installation badge as well.

You still need to explain what is Yunohost and what are the limitations if any. What benefits people have to install Kanboard with Yunohost. Where the users can get help concerning the Yunohost package. Most people have no clue what Yunohost is, so writing few lines doesn't hurt.

PR #3941, opened 2018-07-03
Use SET NAMES instead of charset for DB connection

kufeiko:
When using charset in buildDsn, not all variables are set on the server side. That leads to MySQL assuming latin1 charset and errors with Cyrillic, for example (I figured out the letter 'И', capital breaks the things).


kufeiko:
Shall I create an issue for this?

PR #3944, opened 2018-07-05
API-function "updateTask"  extended

waldemarkurz88:
The time_estimated and time_spent can be updatet via updateTask API call


app/Api/Procedure/TaskProcedure.php
```diff
@@ -166,6 +167,8 @@ public function updateTask($id, $title = null, $color_id = null, $owner_id = nul
             'priority' => $priority,
             'tags' => $tags,
             'date_started' => $date_started,
+			'time_estimated' => $time_estimated,
```
fguillot: This is a minor detail, but can you keep the code indentation consistent with the rest.

Thanks

Issue #3942, opened 2018-07-03
create search within a range of dates


marcihun:
Is it possible, to make a search within a range of dates, the way github allows you to do it?
Like:  "due: 2016-04-30..2016-07-04"

https://help.github.com/articles/understanding-the-search-syntax/#query-for-values-between-a-range

thanks

gerald2545:
is it planned to allow the search within a range for due dates?
thank you

marcihun:
Yes, it is. 
thank you

gerald2545:
cool!!

PR #3948, opened 2018-07-11
Search within a range of dates for creation/moved date field

ksciana:
Solution for #3942


app/Filter/BaseDateRangeFilter.php
```diff
@@ -0,0 +1,54 @@
...
+     * @access protected
+     * @param  string $field
+     */
+    protected function applyDateFilter($field)
+    {
+        $dates = explode('..', $this->value);
+        
+        if(count($dates)=== 2){          
+            $timestampFrom = $this->dateParser->getTimestampFromIsoFormat($dates[0]);
+            $timestampTo = $this->dateParser->getTimestampFromIsoFormat($dates[1]);
```
fguillot: To support user defined date format and ISO format, you should use the method `getTimestamp()` from the `DateParser` class instead.

In this way, `createdRange:"2017-12-09..2017-12-15"` and `createdRange:"09/12/2017..15/12/2017"` should work.

ksciana: OK, but in ``BaseDateFilter`` class is used ``getTimestampFromIsoFormat()`` method form the ``DateParser`` class. It was reference for me.

ksciana:
New changes waiting for review.

letompouce:
@ksciana out of curiosity, why didn't you extend this to the Due date as well?

PR #3984, opened 2018-08-22
Make sure CLI works when called from any directory.

grayed:
Otherwise at least "cli worker" fails to start, when run from
directory different from Kanboard installation one.
```
/home/pers$ doas -u owns_www /var/www/kanboard/cli worker

In QueueManager.php line 67:

  No queue driver defined or unable to connect to broker!


worker

/home/pers$ cd /var/www/kanboard
/var/www/kanboard$ doas -u owns_www ./cli worker

<< succeed >>
```
I had to add this hack to make "cli worker" start from OpenBSD rc.d – the latter forces initial working directory to be same as home directory of user service being run as.

If the proposed solution is not a proper one, please, tell me where to dig for a real fix.

### DIFF
--- a/cli
+++ b/cli
@@ -9,6 +9,7 @@ if (php_sapi_name() !== 'cli') {
 }
 
 try {
+    chdir(__DIR__);
     require __DIR__.'/app/common.php';
     $container['dispatcher']->dispatch('app.bootstrap', new Event);
     $container['cli']->run();


grayed:
Aha! So it's the culprit is the config.default.php, which has the following:

```php
define('PLUGINS_DIR', 'plugins');
```
I copied it to config.php but didn't touch PLUGINS_DIR. After change the line above to:
```php
define('PLUGINS_DIR', __DIR__.DIRECTORY_SEPARATOR.'plugins');
```
the problem went away. Do you want a pull request for that?

fguillot:
Patching `config.default.php` seems to be a better fix than the initial proposition.

grayed:
I'll open new pull request ASAP then.

Issue #3566, opened 2017-11-23
Auto-link duplicates - Feature request


LumpyNorman:
Hello,

Could the act of duplicating a task result in an internal link being applied to the two instances?  This could be an option selected on a checkbox at the time of duplication.

### Actual behaviour
When a task is duplicated, the new item is created but exists as an entirely separate entity.  Regularly the task is generated in a project, but then is assigned to a separate department's Kanboard.  The two instances must then be manually linked to show they are connected. 

### Expected behaviour
When creating a duplicate, in the dialogue window have a checkbox to "Link the duplicates" with the "is duplicated by" setting.

alefaucheur:
+1
Automatic Action request when creating a new task from an existing **sub task**.

**Actual behaviour**
A new task is created in the first column (wherever the "parent" task column was)

**Expected behaviour**
A new task is created in the first column AND a link is made to the parent task (is a child of)

If a template can be provided to handle task creation from a subtask, I believe creating the plugins is very easy.

Thank you ahead of time

fguillot:
@LumpyNorman This is now implemented with pull-request #4047

PR #4048, opened 2018-10-14
Auto link tasks created from a subtask

michaelvickersuk:
Automatically create links for tasks converted from a sub-task as per https://github.com/kanboard/kanboard/issues/3566#issuecomment-351725316

Links are created by default rather than having a checkbox for them to be created. This is so users may easily "undo" accidental creation (by removing the link) or forgetting to check the box (it's created by default). I feel this is the best fit for Kanboard's simplicity and efficiency philosophy.


fguillot:
@michaelvickersuk Thanks! You're on fire :)

PR #4062, opened 2018-10-31
Fix postgres schema version + move to separate lines

trendspotter:
This PR fixes *postgres.sql* schema version which wasn't updated when the schema has been changed in the past. Currently it leads to the following failure during fresh installation:
```
Internal Error: Unable to run SQL migrations: Running migration \Schema\version_109, Running migration \Schema\version_110, SQLSTATE[42701]: Duplicate column: 7 ERROR: column "color_id" of relation "project_has_categories" already exists (You may have to fix it manually)
```

It also puts the version number to a separate line, so it's not that easily overlooked in the future.

- [x] I read the [contributor guidelines](https://docs.kanboard.org/en/latest/developer_guide/contributing.html#i-want-to-contribute-to-the-code)


fguillot:
FYI, this is generated by this Makefile: https://github.com/kanboard/kanboard/blob/master/Makefile#L67-L71 (You may want to add the new line there too).

PR #4069, opened 2018-11-13
Gitlab integration question

cinek810:
I feel like this is bad appproach - but.. I'm not sure how I can address it correctly. Please advice. 
The goal is to assign kanboard task owner based on assignee_id from gitlab. 
[ x ] I read the [contributor guidelines](https://docs.kanboard.org/en/latest/developer_guide/contributing.html#i-want-to-contribute-to-the-code)

### DIFF
--- a/app/Action/TaskCreation.php
+++ b/app/Action/TaskCreation.php
@@ -67,11 +67,20 @@ class TaskCreation extends Base
      */
     public function doAction(array $data)
     {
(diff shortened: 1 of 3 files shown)


fguillot:
You can create a new one specifically for GitLab. Or, you can override the existing one. Create a class that inherit from this one and override the method  `doAction()`, then register your action.

PR #4097, opened 2018-12-10
Add white color

mirabellette:
I wanted to have a white box task only separated by colors tag but the white colors wasn't available.

This pull request makes the white color available. If you look carefully, you could see the border is not black but a little bit lighter. The black was too flashy compare to the other one.

[] I read the [contributor guidelines](https://docs.kanboard.org/en/latest/developer_guide/contributing.html#i-want-to-contribute-to-the-code)

### DIFF
--- a/app/Template/category/create.php
+++ b/app/Template/category/create.php
@@ -8,7 +8,7 @@
     <?= $this->form->text('name', $values, $errors, array('autofocus', 'required')) ?>
 
     <?= $this->form->label(t('Color'), 'color_id') ?>
-    <?= $this->form->select('color_id', array('' => t('No color')) + $colors, $values, $errors, array(), 'color-picker') ?>
+    <?= $this->form->select('color_id', $colors, $values, $errors, array(), 'color-picker') ?>
 
     <?= $this->modal->submitButtons() ?>
 </form>
--- a/tests/units/Model/ColorModelTest.php
+++ b/tests/units/Model/ColorModelTest.php
@@ -41,11 +41,11 @@ class ColorModelTest extends Base
         $colorModel = new ColorModel($this->container);
 
         $colors = $colorModel->getList();
-        $this->assertCount(16, $colors);
+        $this->assertCount(17, $colors);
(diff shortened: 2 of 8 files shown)


mirabellette:
No problem,

Actually, there is a "no color" color which is equivalent to the white color I created but it is not displayed in the task creation when you have the possibility to choose the color. 

I will investigate a little bit more about that.

app/Template/category/create.php
```diff
@@ -8,7 +8,7 @@
     <?= $this->form->text('name', $values, $errors, array('autofocus', 'required')) ?>
 
     <?= $this->form->label(t('Color'), 'color_id') ?>
-    <?= $this->form->select('color_id', array('' => t('No color')) + $colors, $values, $errors, array(), 'color-picker') ?>
```
fguillot: You have to keep the default "No color", otherwise this is going to brake the existing behaviour for existing categories without colour.

tests/units/Model/ColorModelTest.php
```diff
@@ -11,8 +11,6 @@ public function testFind()
     {
         $colorModel = new ColorModel($this->container);
         $this->assertEquals('yellow', $colorModel->find('yellow'));
-        $this->assertEquals('yellow', $colorModel->find('Yellow'));
-        $this->assertEquals('dark_grey', $colorModel->find('Dark Grey'));
```
fguillot: Why do you remove these assertions?

mirabellette: These assertions are a duplicate of code.

app/Template/category/create.php
mirabellette: What do you mean by the existing behavior? The CSS is still well rendered with the modification.

tests/units/Model/ColorModelTest.php
fguillot: This is not a duplicate. Pay attention to the character case.

fguillot:
Closing stale pull-request after 1 month of inactivity.

Issue #4100, opened 2018-12-11
Closed tasks should be displayed by default in boards and dashboard


vincowl:
### Actual behaviour

- In a project's board, only opened tasks are displayed by default
- In a project's board, default filter resets filters to open tasks 
- In a project's board, 'My tasks' shows only open tasks
- In the dashboard, closed tasks are not counted.

### Expected behaviour
- In a project's board, opened and closed tasks should be displayed by default in order to have a complete view. Pre-existing filters are already defined to filter open or closed tasks
- In a project's board, default filter should resets filters to open and closed tasks to be compliant to the above statement.
- In a project's board, 'My tasks' should show open and closed tasks. This leads the previous statement too. It is due to the fact that I use several columns ("New", "To be processed", "In progress", "Done"). The "Done" column should contain closed tasks.
- In the dashboard, closed tasks should be added to open ones (especially in the previous "Done" column)

### Configuration

- Kanboard version: 1.2.6
- Database type and version: sqlite 3.24.0
- PHP version: 7.2.10
- OS: MacOS Mojave (10.14.1) 
- Browser: Safari 12.0.1

creecros:
Personally, I would disagree with this.

vincowl:
So, why not adding a "status:all" default filter and an option in kanboard's setup to choose a default filter to apply instead of forcing the actual behavior ? Several issues seems to have a close need as this one : #4076, #3694, #3581, #3413, #3225
Maybe a bit more explanation on this point of view ?

creecros:
Sorry, not meaning offense, just a point of view. 

My point of view: Closed tasks are closed, so they should be hidden, by default.

creecros:
I agree with choice. Just not as a default. I could go through a lot of reason as to why that would suck, "for me", but better to just say it would pose an issue.

rafacamargo123:
I thing this should be configurable, then. We use one project per team and we're currently running for over 2 years. We have exactly 2648 closed tasks and showing all tasks takes several seconds. Maybe putting this setting as a constant in config.php would be easier to implement and will solve your issue without messing with everyone's kanboard

creecros:
probably simpler to just use the configModel.

loonies:
Why are you closing task if you want them to remain visible i.e. open? Doesn't make sense.

vincowl:
[...]

PR #4101, opened 2018-12-11
Closed tasks should be displayed by default in boards and dashboard

vincowl:
Cf Issue #4100 

### DIFF
--- a/app/Core/User/UserSession.php
+++ b/app/Core/User/UserSession.php
@@ -223,7 +223,7 @@ class UserSession extends Base
     public function getFilters($projectID)
     {
         if (! session_exists('filters:'.$projectID)) {
-            return session_get('user')['filter'] ?: 'status:open';
+            return session_get('user')['filter'] ?: 'status:all';
         }
 
         return session_get('filters:'.$projectID);
--- a/app/Template/app/filters_helper.php
+++ b/app/Template/app/filters_helper.php
@@ -3,7 +3,7 @@
     <a href="#" class="dropdown-menu dropdown-menu-link-icon" title="<?= t('Default filters') ?>"><i class="fa fa-filter fa-fw"></i><i class="fa fa-caret-down"></i></a>
     <ul>
         <li><a href="#" class="filter-helper filter-reset" data-filter="<?= isset($reset) ? $reset : '' ?>" title="<?= t('Keyboard shortcut: "%s"', 'r') ?>"><?= t('Reset filters') ?></a></li>
-        <li><a href="#" class="filter-helper" data-filter="status:open assignee:me"><?= t('My tasks') ?></a></li>
+        <li><a href="#" class="filter-helper" data-filter="status:all assignee:me"><?= t('My tasks') ?></a></li>
         <li><a href="#" class="filter-helper" data-filter="status:open assignee:me due:tomorrow"><?= t('My tasks due tomorrow') ?></a></li>
         <li><a href="#" class="filter-helper" data-filter="status:open due:today"><?= t('Tasks due today') ?></a></li>
         <li><a href="#" class="filter-helper" data-filter="status:open due:tomorrow"><?= t('Tasks due tomorrow') ?></a></li>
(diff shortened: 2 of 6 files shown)


vincowl:
Should update `UserSessionTest::testFilters` in `kanboard/kanboard/tests/units/Core/User/UserSessionTest.php:86`
```
--- Expected
+++ Actual
@@ @@
-'status:open'
+'status:all'
```
Waiting for your point of view on this PR

rafacamargo123:
Sorry to ask, but how many closed tasks do you have?
My project has over 2000 closed tasks ans this behaviour would break usability (browser takes several seconds to render all tasks)

rafacamargo123:
Sorry to ask, but how many tasks does your project has?
We have a project with over 2000 tasks done, this behaviour would break kanboards usability, because browsers take many seconds to render all the tasks... I think a better solution would be to make the default filter configurable, at least globally, by instalation, maybe (config.php).

vincowl:
I don't have so much tasks indeed. For such large projects, I use Gitlab.
Just for my personal info, do the use of swimlanes reduces tasks' rendering
?

Finally, it seems that everyone agrees to say that this optional default
filter may be a good option. Would it be better to set it up at a global
level or at the project's one ?

Le mar. 11 déc. 2018 à 20:46, rafacamargo123 <notifications@github.com> a
écrit :

rafacamargo123:
I have a swimlane with 127 tasks total and another with over 2500. Keeping only the smaller swimlane opened does not reduce tasks rendering time.
I guess it is easier to implement at global level (can be a simple new constant on config.php), if it suits your needs, but I don't know what most users expect.

vincowl:
Ok. Thanks for this info. I didn't have a look at the page's structure.

Implementation at global level would be perfect to me since all my projects
in this Kanboard instance follow the same guidelines. But that would be a
pain if projects are notre consistent in a single Kanboard instance. If
this global variable is not set, it should of course behave as is.

To have an idea of my perds, check the related PR (note the point that
closed tasks should then be added to open ones in dashboard view). I just
found out that I forgot to modify the tasks count in a project activity.

Le 11 déc. 2018 21:16, "rafacamargo123" <notifications@github.com> a écrit :

I have a swimlane with 127 tasks total and another with over 2500. Keeping
only the smaller swimlane opened does not reduce tasks rendering time.
I guess it is easier to implement at global level (can be a simple new
constant on config.php), if it suits your needs, but I don't know what most
users expect.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/kanboard/kanboard/pull/4101#issuecomment-446346428>,
or mute
the thread

vincowl:
@fguillot I have already done it. But as said in #4100, this project organization also implies to modify the overview dashboard to also count closed tasks in columns, and it means to change the default behavior, not just a way to change the filters. That's why @rafacamargo123's proposal seems ways better to me than simply just add a custom filter.
If default or unset value implies to keep the actual behavior then it won't have any impact on people

creecros:
@vincowl 

You could use use the configModel and add a default filter option in the application settings. In template/config/application.php add:
```php
    <fieldset>
        <?= $this->form->label(t('Default Filter'), 'application_stylesheet') ?>
        <?= $this->form->textarea('application_default_filter', $values, $errors) ?>
    </fieldset>
```
in UserSession.php, change the return in getFilters to:
```php
            return session_get('user')['filter'] ?: $this->configModel->get('application_default_filter', 'status:open');	
```

etc...

could easily be a extension without the need for a PR as well.

vincowl:
@creecros 
Thanks a lot for these infos. I had no time before, but I will have a look at it in the next days. Indeed the solution is to design a plugin. I am lacking infos on how to do it and more precisely on what can be made or not. As I tried to explain, the need is not only limited to the filter aspect but also to the analysis part (I mean include closed tasks to statistics given on the analysis project's mode, but also on dashboards and overviews). How far can a plugin modify the actual behavior ? If this can be done, I will have a look on how to design a plugin.  Thanks !!

creecros:
@vincowl From a plug in perspective, I don't think I've yet to not figure out how to do something...given enough time. I am not working on anything new atm, just maintaining my current repos, so happy to help if needed.

fguillot:
Closing this PR. Sorry, I don't want to change the default behaviour. You can continue the discussion on the forum: https://kanboard.discourse.group/.

Issue #3778, opened 2018-03-20
Automatic actions: Import with "category = no category" not possible


m4dm03:
When you want to use another board as a template or you want to import automatic actions from another board, it is not possible to import automatic actions, where the trigger ist "category = no category"

### Actual behaviour

automatic actions with said trigger will not be imported

### Expected behaviour

automatic actions with said trigger should be imported
### Steps to reproduce
- create automatic action with trigger "category = no category"
- try to import it to another board or try to create a new board with previous board as template

### Screenshots
![kanboard_issue](https://user-images.githubusercontent.com/16886798/37670414-859e283e-2c69-11e8-9e71-aa4aaf538d4a.png)

### Configuration

- Application version: 1.2.0
- PHP version: 7.0.27-1+ubuntu16.04.1+deb.sury.org+1
- PHP SAPI: fpm-fcgi
- OS version: Linux 4.4.0-109-generic
- Database driver: mysql
- Database version: 10.0.34-MariaDB-0ubuntu0.16.04.1
- Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) 
 Chrome/64.0.3282.186 Safari/537.36

PR #4102, opened 2018-12-11
Fix importing of automatic actions

rafacamargo123:
Actions containing "unassigned" or "no category" parametes were not imported

Issue #3778 

### DIFF
--- a/app/Model/ActionParameterModel.php
+++ b/app/Model/ActionParameterModel.php
@@ -147,6 +147,8 @@ class ActionParameterModel extends Base
             case 'project_id':
                 return $value != $project_id ? $value : false;
             case 'category_id':
+                if ($value == 0)
+                    return 0;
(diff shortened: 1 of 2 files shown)


rafacamargo123:
Ok I’ll create the tests. Sorry about the braces.

rafacamargo123:
@fguillot I've changed one test that already existed, adding a new action with a category_id = 0 paramenter and created another test with an action that takes an owner_id as parameter. This test creates 2 actions, one to a valid user and another to "unassigned".
I wasn't sure if I could make this way or if I should've made separated test for each scenario, like

- testCloneProjectWithActionTaskAssignSpecificUser
- testCloneProjectWithActionTaskAssignSpecificUserUnassigned
- testCloneProjectWithActionTaskAssignColorCategoryNoCategory

But I think those names would not make sense because the tests are named based on the action class name...

Please tell me what do you think is better and if necessary I change the code.

fguillot:
This is good enough. Thank you.

Issue #4121, opened 2019-01-10
User can still assign tasks when "Changing assignee is not permitted" role is applied


rafacamargo123:
### Actual behaviour

When a custom role with the Project Restriction "Changing assignee is not permitted" is applied to a user, it can still change tasks assignee using "Assign to me" feature

### Expected behaviour

When the project restriction "Changing assignee is not permitted" is applied to a user, it should not be able to change tasks assignee

### Steps to reproduce

1. create a custom role
1. add the project restriction "Changing assignee is not permitted" to this custom role
1. Assign this custom role to a user
1. Open board view, choose a task which current assignee is not the current user and open this task's popover menu
1. Click on Assign to me

### Configuration

- Kanboard version: 1.2.6
- Database type and version: 5.5.60-MariaDB
- PHP version: 7.0.33
- OS: CentOS 7
- Browser: any

PR #4122, opened 2019-01-10
Fix permission check before "Assign to me"

rafacamargo123:
Users who should not be able to change assignee cannot "assign to me"


rafacamargo123:
This PR fixes #4121

PR #4124, opened 2019-01-10
Hide edit button when user cannot edit task

### LINKED ISSUE #4123: Edit button is shown even if user cannot edit task because of custom role
### Actual behaviour

If you assign a role with the restriction "Project -> Update only assigned tasks is permitted" the edit button is shown on tasks owned by other users.

When the button is clicked, though, nothing happens. Checking the browser inspector you can see failed (403 Unauthorized) XHR requests.

### Expected behaviour

The button shall not be shown if the user has no rights to edit the task

### Steps to reproduce

1. Create a custom role
1. Assign the rule "Project  Update only assigned tasks is permitted" to this role
1. Assign this role to an user
1. Open board containing tasks assigned to another users (or unassigned)
1. Edit button is shown for every task

### Configuration

- Kanboard version: 1.2.6
- Database type and version: 5.5.60-MariaDB
- PHP version: 7.0.33
- OS: CentOS 7
- Browser: any

rafacamargo123:
Check for custom roles to show edit button
Fixes #4123

### DIFF
--- a/app/Template/board/task_private.php
+++ b/app/Template/board/task_private.php
@@ -19,7 +19,9 @@
             <div class="task-board-saving-icon" style="display: none;"><i class="fa fa-spinner fa-pulse"></i></div>
             <?php if ($this->user->hasProjectAccess('TaskModificationController', 'edit', $task['project_id'])): ?>
                 <?= $this->render('task/dropdown', array('task' => $task, 'redirect' => 'board')) ?>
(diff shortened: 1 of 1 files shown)


rafacamargo123:
I've added a new "if" instead of replacing the top "if" because the dropdown menu is still useful even with this project restriction. The dropdown shows other operations such as comment, add links, etc. that are not restricted by that custom role.

The "edit" option inside the dropdown on https://github.com/kanboard/kanboard/blob/0deaeb58e60097b7f0882a05481253ff99a5a65a/app/Template/task/dropdown.php#L4 is already hidden by this same permission check

PR #4139, opened 2019-02-12
Fix &add Translation in Japanese.

sambar-fgfs:
I'm fix wrong translation to Japanese, and add some it.

fix: below line not show correctly, due to not match variable.
'%d task(s) in the column "%s" and the swimlane "%s" will be closed.' => 'カラム「%s」とスイムレーン 「%s」にある%d件のタスクを終了します',
if, 
- column name = "column"
- swimlane name ='swimlane'
- Number of tasks = 3

shown:
"カラム「3」とスイムレーン「column」にある0件のタスクを終了します"


fguillot:
Looks like there is a typo that prevent the build to pass: `Incorrect number of %d in ja_JP translation of: %s moved the task #%d to the column "%s"`

https://travis-ci.org/kanboard/kanboard/jobs/492231395

sambar-fgfs:
I was found mixed '＃%d' and '#%d' in app/Locale/ja_JP/Translations.php .
Please tell me which one I should choose. (or if check passed, I am not touch.)

Issue #4137, opened 2019-02-06
Task relationships are broken on MariaDB instances where auto-increment is != 1


psy-q:
### Actual behaviour

Internal links between tasks either can't be assigned and throw an error "The exact same link already exists" or show up with a nonsense relationship (e.g. "blocks" will show up as "blocks" on the opposite side, instead of "is blocked by").

Only "relates to" actually works.

### Expected behaviour

Internal links can be assigned and show their appropriate labels on each side.

### Steps to reproduce

  - Install Kanboard and use a MySQL/MariaDB instance where auto-increment size is 2
  - Create a new task titled "Task 1"
  - Create a new task titled "Task 2"
  - Try to assign internal link  "Task 1 is a milestone of Task 2"
  - Observe that you get an error "The exact same link already exists" even though none exists
  - Assign internal link "Task 1 blocks Task 2"
  - Observe that Task 2 is not showing "is blocked by Task 1", but paradoxically shows "Task 1 blocks Task 2" and "Task 2 blocks Task 1"

The table with the increments that Kanboard was probably not expecting looks like:

```
+----+-------------------+-------------+
| id | label             | opposite_id |
+----+-------------------+-------------+
|  1 | relates to        |           0 |
|  3 | blocks            |           3 |
|  5 | is blocked by     |           2 |
|  7 | duplicates        |           5 |
|  9 | is duplicated by  |           4 |
| 11 | is a child of     |           7 |
| 13 | is a parent of    |           6 |
| 15 | targets milestone |           9 |
| 17 | is a milestone of |           8 |
| 19 | fixes             |          11 |
| 21 | is fixed by       |          10 |
+----+-------------------+-------------+
```

[Increment sizes in MySQL/MariaDB clusters are often 2, 4 or higher numbers depending on the cluster size](https://mariadb.org/auto-increments-in-galera/), so expecting the IDs to always increment by 1 and [hardcoding them](https://github.com/kanboard/kanboard/blob/master/app/Schema/Mysql.php#L1071) will break Kanboard's relationships on those clusters.

### Configuration

  - Application version: 1.2.8
  - PHP version: 7.0.33-0+deb9u1 
  - PHP SAPI: apache2handler
  - OS version: Linux 4.9.0-8-amd64
  - Database driver: mysql
  - Database version: 10.1.37-MariaDB-0+deb9u1
  - Browser: Mozilla/5.0 (X11; Linux x86_64; rv:65.0) Gecko/20100101 Firefox/65.0

[...]

PR #4140, opened 2019-02-13
Fix internal link ids (e.g. for MariaDB clusters with auto_increment_increment != 1)

psy-q:
This should take care of #4137. 

Apologies for the messy commits, if this is not OK I can try to somehow make it cleaner, though I'm not sure how right now (not sure how my earlier PR was merged and why my old commit still shows).

### DIFF
--- a/app/Schema/Mysql.php
+++ b/app/Schema/Mysql.php
@@ -1069,17 +1069,32 @@ function version_46(PDO $pdo)
     $pdo->exec("CREATE UNIQUE INDEX task_has_links_unique ON task_has_links(link_id, task_id, opposite_task_id)");
 
     $rq = $pdo->prepare('INSERT INTO links (label, opposite_id) VALUES (?, ?)');
+    
+    # ID cannot be known at time of record creation so we have to update it after the fact
+    # On MariaDB clusters auto-increment size is normally != 1, so relying on increments of 1 would break
+    $arq = $pdo->prepare('UPDATE links SET opposite_id=? WHERE label=?');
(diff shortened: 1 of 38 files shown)


app/Schema/Mysql.php
```diff
@@ -1069,17 +1069,32 @@ function version_46(PDO $pdo)
...
-    $rq->execute(array('is duplicated by', 4));
-    $rq->execute(array('is a child of', 7));
-    $rq->execute(array('is a parent of', 6));
-    $rq->execute(array('targets milestone', 9));
-    $rq->execute(array('is a milestone of', 8));
-    $rq->execute(array('fixes', 11));
-    $rq->execute(array('is fixed by', 10));
+
+    $rq->execute(array('blocks', 0));
+    $rq->execute(array('is blocked by', get_last_insert_id($pdo)));
```
fguillot: From where is coming the function `get_last_insert_id()`?

psy-q: I've seen it used elsewhere: https://github.com/kanboard/kanboard/blob/master/app/Schema/Migration.php#L70

fguillot: Ah ok. I don't even remember what I did :)

Give me some time to test your pull-request on a Galera cluster.

psy-q: Great, thanks a lot! We tested it on an MMM cluster, but Galera is better, it automatically changes the `auto_increment_increment` as nodes join and leave the cluster.

PR #4158, opened 2019-03-12
it_IT translation update

marcosegato:


fguillot:
The build doesn't pass because there is a typo: `Incorrect number of %d in it_IT translation of: %d invitation was sent.
Failed asserting that 0 matches expected 1.`

PR #4177, opened 2019-04-13
Add LDAP nested groups processing for users

Saruspete:
Provide recusive processing of groups member of groups, only when
LDAP_GROUP_USER_FILTER is provided.

Example: Group1( Group2( User ) )

With only direct membership proessing, user is only member of Group2
With recursion, User will be member of Group2 and Group1 too.

Discuss on https://kanboard.discourse.group/t/ldap-nested-groups/319

### DIFF
--- a/app/Core/Ldap/User.php
+++ b/app/Core/Ldap/User.php
@@ -88,16 +88,34 @@ class User
      * @param  string  $username
      * @return string[]
      */
+    protected function getGroupsRecursive(Entry $entry, $member, &$groupsAll)
+    {
+
+        $groups = $this->group->find(sprintf($this->getGroupUserFilter(), $member));
+
+        foreach ($groups as $group) {
+            $groupId = $group->getExternalId();
+
+            if (!isset($groupsAll[$groupId])) {
+                $groupsAll[$groupId] = $member;
+                $this->getGroupsRecursive($entry, $groupId, $groupsAll);
+            }
+        }
+    }
+
     protected function getGroups(Entry $entry, $username)
     {
         $groupIds = array();
 
+        // If LDAP_GROUP_FILTER has been filled, use it to get groups referring to user
         if (! empty($username) && $this->group !== null && $this->hasGroupUserFilter()) {
-            $groups = $this->group->find(sprintf($this->getGroupUserFilter(), $username));
+            // Try with the dn (eg: member=uid=username,ou=accounts,dc=local)
+            $this->getGroupsRecursive($entry, $entry->getDn(), $groupIds);
+            // try with the username (eg: memberUid=username)
+            $this->getGroupsRecursive($entry, $username, $groupIds);
 
-            foreach ($groups as $group) {
-                $groupIds[] = $group->getExternalId();
-            }
+            $groupIds = array_keys($groupIds);
+
         } else {
             $groupIds = $entry->getAll($this->getAttributeGroup());
         }
@@ -134,7 +152,7 @@ class User
 
             if ($groupId === strtolower($this->getGroupManagerDn())) {
                 // Intermediate role found : we must continue to loop, maybe admin role after ?
-	        $role = Role::APP_MANAGER;
+                $role = Role::APP_MANAGER;
             }
         }
 
--- a/app/Core/Ldap/User.php
+++ b/app/Core/Ldap/User.php
@@ -78,15 +78,12 @@ class User
     }
 
     /**
-     * Get user groupIds (DN)
-     *
-     * 1) If configured, use memberUid and posixGroup
-     * 2) Otherwise, use memberOf
+     * Fill the 3rd param with all memberships of the 2nd param
      *
      * @access protected
-     * @param  Entry   $entry
-     * @param  string  $username
-     * @return string[]
+     * @param  Entry     $entry
+     * @param  string    $member
+     * @param  string[]  $groupsAll
      */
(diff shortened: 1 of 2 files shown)


app/Core/Ldap/User.php
```diff
@@ -88,16 +88,34 @@ public function find($query)
...
     {
         $groupIds = array();
 
+        // If LDAP_GROUP_FILTER has been filled, use it to get groups referring to user
         if (! empty($username) && $this->group !== null && $this->hasGroupUserFilter()) {
-            $groups = $this->group->find(sprintf($this->getGroupUserFilter(), $username));
+            // Try with the dn (eg: member=uid=username,ou=accounts,dc=local)
+            $this->getGroupsRecursive($entry, $entry->getDn(), $groupIds);
+            // try with the username (eg: memberUid=username)
+            $this->getGroupsRecursive($entry, $username, $groupIds);
```
fguillot: Why do you need to try with the DN and then username?

app/Core/Ldap/User.php
```diff
@@ -88,16 +88,34 @@ public function find($query)
...
+    {
+
+        $groups = $this->group->find(sprintf($this->getGroupUserFilter(), $member));
+
+        foreach ($groups as $group) {
+            $groupId = $group->getExternalId();
+
+            if (!isset($groupsAll[$groupId])) {
+                $groupsAll[$groupId] = $member;
+                $this->getGroupsRecursive($entry, $groupId, $groupsAll);
```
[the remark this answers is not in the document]
Saruspete: The use of the isset( groupsAll ) + reference passing is there to avoid processing a group that has already been processed (or is in progress, as registered before the recursion call) thus avoiding infinite recursion.

That would be nice to do a test with it nonetheless, but I don't know phpunit and the current tests well enough to do it atm.

app/Core/Ldap/User.php
Saruspete: Some implementations are done with a weak reference to the name, like (posixGroup + memberuid = username), while some other are done with the strong reference of the DN (groupOfNames + member = uid=username,ou=people,dc=local).
This strong reference is sometimes prefered, as the overlay can update all references when the targeted object is renamed.

Before, it was only done with the username, thus only working with weak references. Now I try with both, as it's easier than letting the user choose in the config, and introducing a new "replacement format", like "%user% or "%dn%".

Saruspete:
I tried to change the test units with multiple ways, but I didn't manage to make phpunit process the 2 calls to find() to get 1 result and 1 another.

I wanted to implement the multi-groups test too, but my skills are lacking there :( 

Any help on this would be welcome ! :)
```
./vendor/bin/phpunit -c tests/units.sqlite.xml  --filter LdapUser 
PHPUnit 5.7.24 by Sebastian Bergmann and contributors.

.....FFF................                                          24 / 24 (100%)

Time: 1.98 seconds, Memory: 34.00MB

There were 3 failures:

1) LdapUserTest::testGetUserWithAdminRoleAndPosixGroups
Expectation failed for method name is equal to <string:find> when invoked zero or more times
Parameter count for invocation Kanboard\Core\Ldap\Group::find('(&(objectClass=posixGroup)(me...ocal))') is too low.

/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:94
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:113
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:172
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:74
/home/www/c/html/public/kanboard/tests/units/Core/Ldap/LdapUserTest.php:557

2) LdapUserTest::testGetUserWithManagerRoleAndPosixGroups
Expectation failed for method name is equal to <string:find> when invoked 1 time(s)
Parameter 0 for invocation Kanboard\Core\Ldap\Group::find('(&(objectClass=posixGroup)(me...ocal))') does not match expected value.
--- Expected
+++ Actual
@@ @@
-'(&(objectClass=posixGroup)(memberUid=my_ldap_user))'
+'(&(objectClass=posixGroup)(memberUid=uid=my_ldap_user,ou=Users,dc=kanboard,dc=local))'

/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:94
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:113
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:172
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:74
/home/www/c/html/public/kanboard/tests/units/Core/Ldap/LdapUserTest.php:664

3) LdapUserTest::testGetUserWithUserRoleAndPosixGroups
Kanboard\Core\Ldap\Group::find('(&(objectClass=posixGroup)(me...ocal))') was not expected to be called more than once.

/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:94
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:101
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:113
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:172
/home/www/c/html/public/kanboard/app/Core/Ldap/User.php:74
/home/www/c/html/public/kanboard/tests/units/Core/Ldap/LdapUserTest.php:775
```

fguillot:
Please, rebase and resubmit your PR. Other LDAP enhancements have been merged and that conflict with your changes.

Sorry for the very long delay.

PR #4183, opened 2019-04-28
[WIP] Add In-Reply-To header to email notifications

fguillot:


lukasjuhrich:
I think this would be a great improvement!

To build on this: wouldn't it be possible to set a reasonable Message-ID? Otherwise, I'm not sure how this is supported with standard mail-clients.

For instance, github uses the following scheme for the `Message-ID` and `In-Reply-To` headers:
- The „root event“ (creation) has a message-ID of the form `<org/project/object_type/object_number@github.com>` (think object_type=issue), so we could use something like `<project-id/task-id@$kanboard_server>`
- Subsequent events set the (deterministic) Root-Event Message-ID Header in `In-Reply-To` and have Message-IDs of the form `<org/project/object_type/object_number/event_type/event_number@github.com>`. I don't think the distinction of event types is necessary here, and would propose to append the timestamp to get a unique value that still clearly relates to the root Message-ID.

fguillot:
Abandoning this PR. I never took the time to finish it.

Issue #4175, opened 2019-04-09
`exec format error` on ARM


ModischFabrications:
### Actual behaviour

Running Kanboard as a docker image on Raspberry Pi 3 (ARMv71) crashes while starting up with `standard_init_linux.go:190: exec user process caused "exec format error"`

### Expected behaviour

A running instance (started from a multiarchitecture build on docker).

### Steps to reproduce

Run `docker run --name kanboard -p 9020:80 -t kanboard/kanboard` on a Raspberry Pi 3.

### Screenshots

--

### Logs
```
pi@modischpi:~ $ docker run --name kanboard -p 9020:80 -t kanboard/kanboard
standard_init_linux.go:190: exec user process caused "exec format error"
```

Architecture:
```
pi@modischpi:~ $ uname -a
Linux modischpi 4.19.32-v7+ #1210 SMP Tue Apr 2 14:07:04 BST 2019 armv7l GNU/Linux
```

### Configuration

- Kanboard version:  `kanboard/kanboard:latest`
- Database type and version: --
- PHP version: --
- OS: --
- Browser: --

ModischFabrications:
This is likely due to a x86 only image on Docker Hub. Look for `multi-arch` as a keyword, this is an official blogpost for it: https://blog.docker.com/2017/11/multi-arch-all-the-things/

fguillot:
There is no official Docker images for ARM architectures. Contributions are welcome. If you want to give a shot, the Dockerfile is there: https://github.com/kanboard/kanboard/blob/master/Dockerfile

PR #4188, opened 2019-05-13
arm32v7 Dockerfile for RPi 3

Alexander-Krause:
This PR introduces the Dockerfile.arm32v7, which is a complete copy of the original Dockerfile except for the first line (base image). 

The built image was tested on my pi with a Nginx-Reverse Proxy. You can find it [here](https://cloud.docker.com/repository/docker/alexanderkrause/kanboard).

https://github.com/kanboard/kanboard/issues/4175

### DIFF
--- a/Dockerfile.arm32v7
+++ b/Dockerfile.arm32v7
@@ -1,0 +1,26 @@
+FROM arm32v7/alpine:3.9
+
+VOLUME /var/www/app/data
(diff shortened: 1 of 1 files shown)


Alexander-Krause:
@fguillot Will you build the image with another tag, e.g., arm32v7, and push it it to the original kanboard Dockerhub repository?

PR #4204, opened 2019-05-29
 migrate HttpClient to cURL

mildis:
Do not require to enable allow_url_fopen anymore
Allow to exclude hosts from being proxyfied
Requires php-curl


mildis:
@fguillot I have issues since I migrate the server to a Debian Strech. Plugin download and OAuth was failing even after I enabled allow_url_fopen and the like. To me using cURL is a more sustainable way to play with HTTP(S) and the like so here is a patch for this.

fguillot:
What about using curl when the PHP extension is available and fallback to the built-in mechanism otherwise?

mildis:
Here you are.
It was not my first bet as it adds a second point of attention if HttpClient ever evolves.

PR #4217, opened 2019-06-19
Fixes icon opacity when hovered

### LINKED ISSUE #4214: Regression: hovering icons on tasks cards on the board does not highlight/opacify them
Running the latest kanboard version, I noticed that for tasks on a board, in the icons row for things like "Relations", "Subtasks", "Comments", "Description", if you hover the mouse over these semi-transparent icons they don't "highlight" (become fully opaque) like they used to before, to indicate that the mouse is correctly over the target, _except_ in the case of the "comments" icon.

I found the opacification on mouseover very useful to know that my mouse is correctly positioned with precision over the target, before the expanded popup preview even shows up.

So, to summarize, on mouseover:
* Relationships icon don't opacify
* Subtasks icon doesn't opacify
* Comments icon DOES opacify, which is nice
* Description icon doesn't opacify

rafacamargo123:
Handle tooltip icons the same ways link icons are handled 
when hovered.
Fixes #4214


rafacamargo123:
This is the new behaviour introduced with this pull request:

ezgif-3-5a9f5ebb4806

PR #4234, opened 2019-07-09
Save images as PNG to allow transparency.

KN4CK3R:
I have changed the file type of the generated thumbnail from jpeg to png. This change allows to display transparency in uploaded avatars or other images.


fguillot:
The mime type for image attachments need to be changed to `image/png` as well: https://github.com/kanboard/kanboard/blob/master/app/Controller/FileViewerController.php#L119

KN4CK3R:
Right, missed that one.

PR #4238, opened 2019-07-15
Hotfix: Added Apache ext-pdo_sqlite requirement 

gustavorps:
To do deploy in Heroku I needed the SQLite extension


fguillot:
pdo_sqlite is not mandatory because people can choose to use other databases like PostgreSQL or MySQL.

If you are planning to run Kanboard on Heroku, you may want to use their Postgres offering since the local storage is ephemeral?

PR #4242, opened 2019-07-16
Ignore vendor directory

gustavorps:
As suggested in Composer docs [Should I commit the dependencies in my vendor directory?](ttps://getcomposer.org/doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md) and to avoid conflicts in services as Heroku the vendor directory has been removed.


fguillot:
See https://docs.kanboard.org/en/1.2.10/developer_guide/faq.html#why-php-vendor-directory-is-committed-into-the-source-tree

PR #4248, opened 2019-07-23
Fix html parsing on markdown editor

### LINKED ISSUE #4243: `</script>` gets lost on task edit form
### Actual behaviour

anything beyond a closing-script tag (like in a javascript-snippet) gets lost on task edit.

### Expected behaviour

everything stays in description

### Steps to reproduce

1. add the following in a task:

```
<script>This is a test</script>
<script>this should stay in the description</script>
```

2. open task edit for - the second line is gone.

### Screenshots

image

image

### Configuration

- Kanboard version: 1.2.10 (docker)
- Database type and version: mysql
- OS: linux
- Browser: chrome 75

rafacamargo123:
Closes #4243


app/Helper/FormHelper.php
```diff
@@ -232,7 +232,7 @@ public function textEditor($name, $values = array(), array $errors = array(), ar
         }
 
         $html = '<div class="js-text-editor" data-params=\''.json_encode($params, JSON_HEX_APOS).'\'>';
-        $html .= '<script type="text/template">'.(isset($values[$name]) ? $values[$name] : '').'</script>';
+        $html .= '<script type="text/template">'.(isset($values[$name]) ? htmlspecialchars($values[$name], ENT_QUOTES, 'UTF-8', true) : '').'</script>';
```
rafacamargo123: I used htmlspecialchars instead of `$this->helper->text->e()` because the `e()` function calls `htmlspecialchars` with `$double_encode = FALSE` and it causes a wrong behaviour when the text contains an already escaped html, like `&amp;`. If you dont double encode id, the editor will load the escaped character `&` instead of the original text.
This only affects the edition, the visualization is still the same: If you type `&gt; &amp;` to a markdown text editor, you will see `> &` on task and comments view.
I tought of changing the `function e`, adding an optional parameter `$double_encode = true`.
If necessary, I'will change the textHelper and use it here.

PR #4250, opened 2019-07-24
Add link button to text editor

rafacamargo123:
Add link button to text editor toolbar


rafacamargo123:
Behaviour:

If no text is selected, then it just adds a `[title](http://...)` to current cursor position and selects nothing:

ezgif-5-d5e550c52bc4

The title portion uses the internationalized string "Title".

If a text starting by http is selected, it will assume it's a link, so will populate the url portion of the markdown link with the selected text and select the "Title" portion so the user can change it:

ezgif-5-3e57c2766c29

If the selected text does not start with http, it's considered a title and the url portion is selected:

ezgif-5-d31f5aedd8b6

strukturart:
the safety starts with the little things ;-)
please add https:// as default protocol

PR #4258, opened 2019-08-03
Add missing text-caret vendor js library

### LINKED ISSUE #4257: text-caret js library is missing from minified js files
Comment mentions are not working as stated by perry on https://kanboard.discourse.group/t/comment-mention-without-suggestion/567/8.

Browser log:

```
app.min.js?1564087401:231 Uncaught ReferenceError: getCaretCoordinates is not defined
    at renderMenu (app.min.js?1564087401:231)
    at onItemFetched (app.min.js?1564087401:224)
    at app.min.js?1564087401:223
    at XMLHttpRequest.request.onreadystatechange (app.min.js?1564087401:29)
```

 I've tracked down the issue to app.min.js and vendor.min.js. Neither of them have the definition of the function getCaretCoordinates

rafacamargo123:
Closes #4257


app/Console/JsCommand.php
```diff
@@ -67,6 +67,7 @@ class JsCommand extends BaseCommand
         'assets/vendor/c3/c3.min.js',
         'assets/vendor/isMobile/isMobile.min.js',
         'assets/vendor/marked/marked.min.js',
+        'assets/vendor/text-caret/index.js',
```
fguillot: Hum.. it's weird, it's included above in `$appFiles`. I will check when I will have 5 minutes.

rafacamargo123: Oops! Missed that. I only checked the vendor because it was inside vendors folder. It’s not working because it’s written “textarea-caret” instead of text-caret. That’s why it is not being included.

rafacamargo123: I've just updated the PR, fixing the reference in `$appFiles`

https://github.com/kanboard/kanboard/pull/4258/files#diff-de951cd95adad4d2ed69fcfdb50460a2L30-L32

rafacamargo123:
I see that you check if the file exists in the `concat_files` function, however the `glob` only returns matching files. I've added two flags to the glob function that will make the `file_exists` check valid.
If a pattern matches no file, the blob function returns the pattern itself, allowing the file_exists function to check it and return false.

fguillot:
Good catch!

I merged only the first 2 commits because I also fixed the Postgresql build in master (I noticed you did the same fix, but too late).

Thanks

PR #4262, opened 2019-08-09
Menu position on header view

dreamcoder8:
Added following hooks to be used to place new menu items 

- view-switcher-before-ProjectOverview
- view-switcher-before-BoardView
- view-switcher-before-TaskList

### DIFF
--- a/app/Template/project_header/views.php
+++ b/app/Template/project_header/views.php
@@ -1,6 +1,6 @@
 <ul class="views">
     
-    <?= $this->hook->render('template:project-header:view-switcher-before-oroject-overview', array('project' => $project, 'filters' => $filters)) ?>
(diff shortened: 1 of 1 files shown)


app/Template/project_header/views.php
```diff
@@ -1,10 +1,19 @@
 <ul class="views">
+    
+    <?= $this->hook->render('template:project-header:view-switcher-before-ProjectOverview', array('project' => $project, 'filters' => $filters)) ?>
```
fguillot: it would probably be better to keep the naming consistent and use something like `template:project-header:view-switcher-before-project-overview` instead of `template:project-header:view-switcher-before-ProjectOverview`.

dreamcoder8: Done.

PR #4265, opened 2019-08-16
hide empty time on card

BlueTeck:
With this PR the due date time is hidden on the card, if the time is 00:00
I think it makes the card and due date more readable.

Forum:
https://kanboard.discourse.group/t/hide-due-date-time-on-card-if-no-time-is-specified/577

Screenshot:
image

### DIFF
--- a/app/Template/board/task_footer.php
+++ b/app/Template/board/task_footer.php
@@ -68,7 +68,11 @@
                 <?php endif ?>
                 ">
                 <i class="fa fa-calendar"></i>
-                <?= $this->dt->datetime($task['date_due']) ?>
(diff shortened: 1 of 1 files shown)


app/Template/board/task_footer.php
```diff
@@ -68,7 +68,11 @@
                 <?php endif ?>
                 ">
                 <i class="fa fa-calendar"></i>
-                <?= $this->dt->datetime($task['date_due']) ?>
+                <?php if (date('H', $task['date_due']) > 0 || date('i', $task['date_due']) > 0 ): ?>
```
fguillot: Can we simplify this condition? for example: `date('Hi', time()) === '0000'`

BlueTeck: seems to be much better, I've changed it :)

Issue #3932, opened 2018-06-25
Mouseover tooltips are not disappearing


soudis:
### Actual behaviour

When I hover of the subtasks or the description icons the subtasks or description appear in a tooltip, but they do not disappear until I move the mouse into the tooltip and then leave it again.

### Expected behaviour

The tooltip should disappear after I move the mouse away from the subtasks icon and not into the tooltip container. Otherwise it gets really annoying sometimes, if I just cross the subtasks icon accidentally by moving the mouse over the screen and suddenly somewhere a tooltip appears and stays indefinitely. 

This also blocks other tooltips from appearing. If I move the mouse to different description or subtasks icon, the tooltip is not appearing until i close the previous one by moving the cursor into the tooltip container and then leave it again.

### Steps to reproduce

1. Open any board containing cards with descriptions
2. Hover over the description icon on one of the cards
3. Move cursor away from the description icon and not into the tooltip
4. Tooltip stays indefinitely, blocking parts of the screen and blocks other tooltips from opening

### Screenshots

![kanboard tooltip](https://user-images.githubusercontent.com/16363411/41843714-87c5b9d2-786e-11e8-874f-59c9bd47af6a.png)

### Logs

no logs

### Configuration

- Kanboard version:  1.2.4
- Database type and version: 10.2.7-MariaDB-10.2.7+maria~jessie
- PHP version: 7.1.17
- OS: Ubuntu 16.04 LTS /  Linux 4.9.0-3-amd64
- Browser: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/60.0

Victor-IT:
v1.2.5
I have the same issue.

Kantime:
I'm complaining about this since 1.2.4
Please, do something !
**I returned to 1.2.3 just because of that issue.**

jschuck:
The issue was introduced with changeset 5996a8ab. The tooltips are not neccessarily opening up beneath the mouse cursor and are only closing on a mouseleave event. Which isn't happening unless you move the mouse over the tooltip.

That's what's causing this issue.

neiaberau:
I have the same issue, seen with Vivalidi Browser, Edge Browser
and I believe, this https://github.com/kanboard/kanboard/issues/3915 is about the same issue

jschuck:
[...]

PR #4270, opened 2019-08-22
Fixed issue of tooltip not disapearing.

eSkiSo:
Tooltip would not disapear if mouse was never on it, if you put mouse on an icon and then moved the mouse out without passing by tooltip, the tooltip would remain active until mouse pass and leave or click somewhere else.
With this update tooltip will be removed if you leave target unless you move the mouse to the tooltip.
Fixes issue: #3932


eSkiSo:
PS. I just changed the source tooltip.js, on release that is minified to app.min.js, which i have not changed on this pull, i believe it is minified on release launch or test start?

Issue #4271, opened 2019-08-23
Drag and drop cards on mobile devices cause page to shrink


Haocen:
### Actual behaviour
When drag and drop cards on mobile device using handler on top left, the page shrink to show more visible area

### Expected behaviour
The page should not shrink

### Steps to reproduce
Drag and drop a card

### Screenshots
![kanboard drag drop](https://user-images.githubusercontent.com/10842569/63606123-5bab4880-c59d-11e9-9936-b8d4098bc3c1.gif)

### Configuration

- Kanboard version:
- Database type and version:
- PHP version:
- OS:
- Browser:

PR #4272, opened 2019-08-23
Disable user scaling to avoid page shrinking when drag&drop on mobile device

Haocen:
Closes https://github.com/kanboard/kanboard/issues/4271


fguillot:
I guess that would fix the issues for Android users then.

Haocen:
Yes it will fix for Android users. If I find a way to fix it for iPhone users I'll investigate, thanks for your understanding.

PR #4275, opened 2019-08-26
Add hash to image url to force browser to update profile when changed

### LINKED ISSUE #4241: User avatars change only on (shift-) reload
### Actual behaviour

A user changes their avatar by uploading a new picture. However, the page still shows the old avatar. Other users who have viewed a page with the old avatar will continue to see the old avatar until their caches expire. By using shift-reload, the new avatar is shown.

A similar issue [was already reported](https://github.com/kanboard/kanboard/issues/3463) but it is still present in the latest Kanboard.

### Expected behaviour

The new avatar image should be shown to all users immediately after uploading.

### Steps to reproduce

  1. Go to profile page, upload an avatar.
  2. Upload a different avatar.
  3. Note that the avatar image fails to change.
  4. Reload the page.
  5. Observe that the avatar is now changed.

### Configuration

   - Application version: 1.2.10
   - PHP version: 7.0.33-0+deb9u3
   - PHP SAPI: apache2handler
   - HTTP Client: cURL
   - OS version: Linux 4.9.0-9-amd64
   - Database driver: mysql
   - Database version: 10.1.38-MariaDB-0+deb9u1
   - Browser: Mozilla/5.0 (X11; Linux x86_64; rv:67.0) Gecko/20100101 Firefox/67.0

rafacamargo123:
Closes #4241


rafacamargo123:
Browser will still use cache. When user uploads a new avatar, the file path will change (it is based on file hash), so the hash on avatar url will change and the browser will download a new file.

Issue #4291, opened 2019-09-13
column limit should not respect filter


BlueTeck:
### Actual behaviour

If there is a higher number of tasks than the task limit in a column, the column is displayed red.

If you filter your board and the number of tasks is reduced, the column isn't red anymore.

see also https://github.com/kanboard/kanboard/issues/2488

### Expected behaviour

The task limit should consider all open tasks in this column, not the number of filtered and displayed tasks.

The number from [Taskfinder->countByColumnId](https://github.com/kanboard/kanboard/blob/master/app/Model/TaskFinderModel.php#L358) should be used.

### Steps to reproduce

Create column with task limit, create more tasks than the limit, filter some tasks.

### Source Code

https://github.com/kanboard/kanboard/blob/master/app/Template/board/table_tasks.php#L6

PR #4292, opened 2019-09-13
check tasklimit against open tasks

BlueTeck:
possible fix for #4291 might not be the best way

### DIFF
--- a/app/Formatter/BoardSwimlaneFormatter.php
+++ b/app/Formatter/BoardSwimlaneFormatter.php
@@ -101,6 +101,8 @@ class BoardSwimlaneFormatter extends BaseFormatter implements FormatterInterface
             foreach ($swimlane['columns'] as $columnIndex => &$column) {
                 $column['column_nb_tasks'] = $this->swimlanes[0]['columns'][$columnIndex]['column_nb_tasks'];
                 $column['column_nb_score'] = $this->swimlanes[0]['columns'][$columnIndex]['column_score'];
+                // add number of open tasks to each column, ignoring the current filter
+                $column['column_nb_open_tasks'] = $this->columns[array_search($column['id'], array_column($this->columns, 'id'))]['nb_open_tasks'];
             }
         }
 
@@ -108,7 +110,7 @@ class BoardSwimlaneFormatter extends BaseFormatter implements FormatterInterface
     }
 
     /**
-     * Calculate stats for each column acrosss all swimlanes
+     * Calculate stats for each column across all swimlanes
      *
      * @access protected
      * @param  array $columns
--- a/app/Formatter/BoardFormatter.php
+++ b/app/Formatter/BoardFormatter.php
@@ -44,7 +44,7 @@ class BoardFormatter extends BaseFormatter implements FormatterInterface
     public function format()
     {
         $swimlanes = $this->swimlaneModel->getAllByStatus($this->projectId, SwimlaneModel::ACTIVE);
-        $columns = $this->columnModel->getAll($this->projectId);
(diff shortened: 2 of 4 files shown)


app/Formatter/BoardSwimlaneFormatter.php
```diff
@@ -101,6 +101,7 @@ public function format()
             foreach ($swimlane['columns'] as $columnIndex => &$column) {
                 $column['column_nb_tasks'] = $this->swimlanes[0]['columns'][$columnIndex]['column_nb_tasks'];
                 $column['column_nb_score'] = $this->swimlanes[0]['columns'][$columnIndex]['column_score'];
+                $column['column_nb_open_tasks'] = $this->taskFinderModel->countByColumnId($column['project_id'], $column['id'] );
```
fguillot: That might generate too many SQL queries for large board. Need to find another way of fetching the number of open tasks.

BlueTeck: @fguillot I've changed the implementation, it should be a single select now.

BlueTeck:
@fguillot I've changed the implementation, it should be a single select now.

app/Formatter/BoardSwimlaneFormatter.php
```diff
@@ -104,19 +105,45 @@ public function format()
...
 
     /**
-     * Calculate stats for each column acrosss all swimlanes
+     * Add number of open tasks to eeach column, ignoring the current filter
+     *
+     * @access protected
+     */
+    protected function addNumberOfOpenTasksInColumn()
+    {
+        $column_open_task = $this->db->table(TaskModel::TABLE)
```
fguillot: This is better but we should avoid doing SQL queries in a Formatter class. This class is about formatting the results from the database.

That would be nice if you could find a way to move this method to the models.

BlueTeck: @fguillot I've changed it again. It now uses an existing method which returns to taskcount.

BlueTeck:
@fguillot I've changed it again. It now uses an existing method which returns to taskcount.

BlueTeck:
@fguillot whats your opinion on this?

app/Formatter/BoardSwimlaneFormatter.php
```diff
@@ -104,19 +104,39 @@ public function format()
...
 
     /**
-     * Calculate stats for each column acrosss all swimlanes
+     * Add number of open tasks to each column, ignoring the current filter
+     *
+     * @access protected
+     */
+    protected function addNumberOfOpenTasksInColumn()
+    {
+        $columns = $this->columnModel->getAllWithTaskCount($this->swimlanes[0]['columns'][0]['project_id']);
```
fguillot: If you could move this call in the function `format()`..

app/Formatter/BoardSwimlaneFormatter.php
```diff
@@ -104,19 +104,39 @@ public function format()
...
+
+        if (!empty($columns)) {
+            foreach ($this->swimlanes as &$swimlane) {
+                foreach ($swimlane['columns'] as $columnIndex => &$column) {
+                    $column['column_nb_open_tasks'] = $columns[array_search($column['id'], array_column($columns, 'id'))]['nb_open_tasks'];
```
fguillot: Is it possible to move this line in the loop above (line 103)? That would avoid doing twice the same loop.

BlueTeck:
@fguillot I refactored it again, I think it even better now. 

Unfortunately the last test failed, but it is not related to this change, just an unlucky timing issue, maybe you could restart the test?

fguillot:
Thanks for addressing my comments.

Issue #4313, opened 2019-10-07
No privacy from admins


JavierYuste:
Why are administrators able to see everything? That makes no sense from a privacy point of view. Private projects are not really private then.

BlueTeck:
I stumbled on it too.

pcadeot:
+1
This point is becoming quite problematic in my company, is there any plan to enable some options in general config to enforce privacy for so called private projects ?

BlueTeck:
I've created PR #4317

JavierYuste:
Thanks a lot for the PR @BlueTeck !

pcadeot:
Very nice ! Thanks ! @BlueTeck

fguillot:
Private projects have been renamed personal projects. They are not "private".

PR #4317, opened 2019-10-11
new config parameter ADMIN_RESPECTS_PRIVACY

BlueTeck:
This PR adds a new config parameter which provides an option, that admins couldn't see every project.

https://github.com/kanboard/kanboard/issues/4313

@fguillot what do you think? There might be a better name for the parameter ;)


fguillot:
The admin role have access to everything intentionally. The purpose of this role is to manage the Kanboard instance. There is another role called "Manager" that might be more suited for daily use.

Private projects are projects with users management disabled. They are not private in the sense where no one can see your projects.

BlueTeck:
In our case the problem is, we need some users who can manage/create users and groups. But they should not be allowed to view/access every project easily.

JavierYuste:
Exactly. I think there is no need for an admin to access private projects in order to manage the Kanboard instance. Managing settings is one thing, accessing projects you should not be allowed to is another completely different thing. "Managing" projects should be done by their "managers", as I understand it.

I think the solution proposed by @BlueTeck is quite nice and would make everyone happy, allowing them to choose whether they want to respect privacy or not.

rafacamargo123:
You must add the constant to [constants.php](/kanboard/kanboard/tree/master/app/constants.php) or it will not be defined.
config.default.php is only an example to the user.
May I suggest to keep the default value to false to prevent changing the current behaviour for everybody?

fguillot:
Then having an intermediate role between Admin and Manager seems better from my point of view. It's explicit and make it easier to document: https://docs.kanboard.org/en/latest/user_guide/users.html#application-roles

pcadeot:
Perhaps "private projects" is indeed confusing if it intentionnally doesn't prevent admin to access content, which translates to intentionnally doesn't mean "private"... 
"personal project" would be better as it is de facto "project without managing users"
"solo project" ? "one man project" ?
:)
perhaps this could lead to a privacy option for a personal project : as a user I create a personal project, and I tick a private box if I'd like "no one not even admins" access my board ?
...
my 2 cts
PC

JavierYuste:
I am afraid I am not able to see your point @fguillot . What would an intermediate role between Admin and Manager be able to do? How would that solve the problem?

jktoole:
I agree with fguillot. An adminstrator should be able to administrate everything, it seems counter-intuitive to add what seems like exceptional/non-standard behaviour for a specific use case. If there was ever a problem with a user's private project that an administrator needed to fix and this setting was enabled, they'd have to temporarily edit the config or try to fix the problem from the backend - and that after having figured out that this setting existed if they were new to the task.

The problem being described is where a user needs more privileges than a manager but less than an administrator, so it should probably be another application role rather than a feature-specific setting. This new role might also be denied other permissions appropriate for its place between manager and administrator.

Until the new role is set up, users that need the feature could use the code from this commit to temporarily patch their systems.

I also agree with pcadeot, 'personal' would be a better description than 'private' for projects with a single user.

JavierYuste:
As long as there is a single role able to see private/personal projects (whatever you wanna call them, and whatever name that role is given) the problem is not fixed. Adding another role is not a fix. The discussion seems to have been like:

Why can admins access everything? No privacy! -> True, lets add an option so users can decide whether to respect privacy or not -> Hey we can add a new role -> Yeah the problem is that "private" is confusing, it should be named "personal"

So the current discussion of privacy vs personal projects and adding a new role have nothing to do with the exposed problem, IMHO. You could always make a distinction and add a third type of project, if you are willing to maintain such difference between private and personal projects.

I do not understand why adding an option to allow privacy is not instantly accepted. It should be the default option in my opinion, but that is out of scope for now. 

What I am trying to express here is that adding a privacy option is a requirement for anyone with a minimum security concern, and not having such an option is a point against the project when evaluating the project to deploy it in a business/professional environment.

justb81:
We had this issue with inviting new users to projects and startet a plugin for this: https://github.com/ipunkt/KanboardProjectInvitation

With that plugin as a project manager you are able to invite new users by mail to your managed project. It doesn't allow further group managing with users, but basic project work is possible.

grayed:
> I do not understand why adding an option to allow privacy is not instantly accepted. It should be the default option in my opinion, but that is out of scope for now.
> 
> What I am trying to express here is that adding a privacy option is a requirement for anyone with a minimum security concern, and not having such an option is a point against the project when evaluating the project to deploy it in a business/professional environment.

If user doesn't trust someone else's (Kanboard) server, he shouldn't put anything private there at all. Admin is by definition the one being able to access anything. Or, maybe, you think that e.g. GitHub admins won't be able to look at your private repos? :)

If you want "private" Kanboard projects become really private, then you'll need to implement some sort of per-project or per-user data encryption, with decrypt happening ONLY on client side, at minimum. Otherwise anyone who can access Kanboard database directly (i.e., administrator) still can do anything with those repos. And even after getting pure client-side data encryption you'll run in a lot of sideway information leaks...

PapeCoding:
The comment of @grayed is totally valid, but there are some instances, where it would be nice to have this feature, as @JavierYuste said.
In my specific case, we are hosting KanBoard as a group of people. Surely the admin can access the database if they want to, but doing it from the Frontend is something else. My boss would like to use it, but definitely refuses to, because he has some "secret"-stuff going on, which even the admins should not have such easy access to. For such occurrences, this option would be nice.

liedekef:
I agree that admins shouldn't be able to see all: my manager uses it for personal stuff (even to evaluate me) and I shouldn't be able to see those. This is only 3 lines of code to add, and one could even set the option to "false" to keep the current behaviour, but I'd very much like this functionality.

Issue #3916, opened 2018-06-16
Fix table view in mobile version


ngosang:
### Actual behaviour
See the screenshot

### Expected behaviour
The project selection combo and the status input width should be responsive. Looks like in the mobile browser the window width is small and the input width is fixed.

### Screenshots
![imagen](https://user-images.githubusercontent.com/10577978/41498207-d254288e-7167-11e8-8acf-7f6c9612181c.png)

### Configuration
    Kanboard version: master
    OS: Kubuntu
    Browser: Firefox 60

josch87:
Problem with filter appears when width <385 px. Please support up to at least 320 px width (Mobile S).

SladePrice:
I ran into this problem too and decided to attempt a fix. I have a potential solution for this in a fork. I've attached a couple screenshots of my proposed method for resolving this. One screenshot shows a scroll bar overlapping the submenu, but I tested this on mobile and the bug doesn't exist on Android. I've tested it in as many locations as I can find so far that use input-addon classes, and will finish extensive testing if this is a good solution.

![image](https://user-images.githubusercontent.com/25444175/68814267-928d7980-0635-11ea-85d4-2d3eb919c00f.png)

![image](https://user-images.githubusercontent.com/25444175/68814337-c7013580-0635-11ea-8ae1-b61bb7fb0acf.png)

![image](https://user-images.githubusercontent.com/25444175/68814356-d3858e00-0635-11ea-96b5-ea5233c42f45.png)

![image](https://user-images.githubusercontent.com/25444175/68814482-337c3480-0636-11ea-9349-c24d472da220.png)

![image](https://user-images.githubusercontent.com/25444175/68814542-5870a780-0636-11ea-8ade-2ba9c283392e.png)

If this is an acceptable resolution, I'll submit a pull request. Thanks!

SladeP:
Okay, I adjusted the CSS so the blocks look better when wrapping

![image](https://user-images.githubusercontent.com/25444175/68888767-cbcaf580-06d8-11ea-9e47-69114a2ee955.png)

![image](https://user-images.githubusercontent.com/25444175/68888809-deddc580-06d8-11ea-8564-9c8dcf92d87b.png)

SladeP:
After checking the above fix with other themes, I found out that Greenwing had a good fix for this, so I adopted that fix for the base Kanboard style. See screenshots below. Compatible down to 320px.

[...]

PR #4337, opened 2019-11-14
Fix issue 3916: width of filter bar in mobile.

SladeP:
I made comments on issue #3916 with screenshots and an explanation. Basically, I adopted the way the Greenwing theme handles the filter bar being too long in mobile and applied it to the default Kanboard style. Here's a screenshot. There are more screenshots in my last comment in #3916.

image


SladeP:
I added a quick 2 commits to fix an issue where submenus would sometimes calculate their left offset as negative and get shoved off-screen. Fixed by adding a little javascript check to convert negative offset values to a default of 15px. I was going to add this as a second pull request because I mistakenly thought it was a bug without my first commit. 

Before

image

After

image

PR #4338, opened 2019-11-14
Fix issue 3916: width of filter bar in mobile.

SladeP:
Sorry, I deleted the original pull request in favor of revising this into a single commit. This is my first open source contribution, forgive me.

I made comments on issue #3916 with screenshots and an explanation. Basically, I adopted the way the Greenwing theme handles the filter bar being too long in mobile and applied it to the default Kanboard style. Here's a screenshot. There are more screenshots in my last comment in #3916.

image

Fixing the mobile filter bar introduced a bug where submenus would get shoved off-screen when calculating where to place them relative to the clicked icon. To fix, I added some javascript logic to the submenus so that if the calculated left offset is negative, it should default to 15px. This is only important in mobile views.

Before

image

After

image


ngosang:
I didn't test it, but any improvement in the mobile version will be nice. Thank you!

PR #4351, opened 2019-11-18
Fix #4289. Moving the data-js-lang attribute to the html lang attribute

### LINKED ISSUE #4289: Chrome recognizes Dutch on German project page
### Actual behaviour
Chrome asks if I want to translate dutch on a the German "projects" page (projects management). 

### Expected behaviour
The page is German. Browsers should recognize that it is German.

### Steps to reproduce
- Change language of user interface to German (or any other language)
- Go to projects management
- Chrome want to translate the page. 

It would be great if any page would have `<html` `lang="de">` in head. 

### Screenshots
image

### Configuration

Version: 1.2.11
PHP Version: 7.1.31
PHP SAPI: apache2handler
HTTP-Client: cURL
OS Version: Linux 3.16.0-7-amd64
Datenbanktreiber: mysql
Datenbank Version: 5.7.24
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.132 Safari/537.36

SladeP:
Fix for #4289. 

Browsers recognize the language of the page by reading the `lang` attribute of the html tag. We already track the javascript language in the body element with the attribute `data-js-lang`. This fix gets rid of the `data-js-lang` attribute and instead places it in `<html lang="<?= $this->app->jsLang() ?>">` It also updates timepicker to find the language in the html element.

Tested all languages and Chrome shows the correct language for each in the translation dropdown.

Except for some reason Vietnemese doesn't send any js language, so it defaults to "en" to the lang attribute. And, actually, looks like the data-js-lang attribute with Vietnemese is also broken and the attribute doesn't even get initialized in the body.

I'll make a new issue for the Vietnemese js-lang and fix it asap.


SladeP:
Closing this and opening a new pull request with a cleaner commit history after merging in master.

PR #4363, opened 2019-12-03
Fix #4213. Adjust width of timetracking column. Make subtasks not wrap under icons.

### LINKED ISSUE #4213: Space savings: subtasks table's first column should have flexible width
Tested with the latest stable version (the bug has been around forever howeve), this is what I typically get in the UI:

kanboard subtask table columns default sizing

This leads to the subtask titles wrapping around in a pretty inelegant way, while the third (time spent / stopwatch) column is full of unused space.

Looking at the HTML code, the table's first "th" element is `<th class="column-45">Title</th>`, the second is `<th class="column-15">Assignee</th>` and the third is just `<th>Time tracking</th>`.

The fix appears to be simple then: remove the class from the first th, and put `class="column-15"` onto the third th. And this is the result:

kanboard subtask table columns proposed sizing

SladeP:
Fixes #4213 

The changes in this commit are expressed in detail in my comment on the issue #4213 . In summary:

* I set the time tracking column to column-30 because that's the smallest size that will accommodate all languages (Spanish being the longest)

* I made the time tracking column next wrap in mobile because we don't need this huge long line for scrolling in mobile, it's readable enough when wrapped

* I grouped the move icon and settings icon and checkbox into flex containers so text would wrap under itself, not under those icons. This makes wrapping look cleaner and easier to read.

Now, screenshots:

# My Subtasks Before

mysubtasks_before

mysubtasks_mobile_before

# My Subtasks After

mysubtasks_after

mysubtasks_mobile_after

# Tasklist Before

tasklist_before

tasklist_mobile_before

# Tasklist After

tasklist_after

tasklist_mobile_after

# Task View Before

taskview_before

taskview_mobile_before1

taskview_mobile_before2

taskview_mobile_before3

# Task View After

taskview_after

taskview_mobile_after1

taskview_mobile_after2

taskview_mobile_after3

# Task View Spanish Before

taskview_spanish_before

# Task View Spanish After

taskview_spanish_after


fguillot:
Looks good except GitHub detected a conflict with `app.min.css`. You might need to update your branch.

SladeP:
Closing because trying to update the branch deleted the old branch so this pull request references a commit that no longer exists. Sorry, still trying to figure out the merges and conflict resolution.

PR #4373, opened 2019-12-13
Avoid implicit line wrapping for input-addon-item CSS style

grayed:
This makes filter buttons look like fine again for me (Chromium on OpenBSD/amd64-CURRENT, 1920x1080 screen running cwm). See attached images for difference:

Before:
before

After:
after


fguillot:
You should also include the app.min.css

grayed:
@fguillot done. I've thought syncing better made separately, but, of course, your app — your rules. :)

fguillot:
I'm trying to reproduce your issue, but I don't see the problem on my side. I tested with the same screen resolution, language set to Russian, and Chromium 79.0.3945.79 (The actual version shipped with Ubuntu). Same on Firefox 71.

What Chromium version do you have on OpenBSD?

I could merge but I don't see the difference after and before...

grayed:
The problem disappeared after updating Calendar plugin to recently released 1.1.1. Thank you for your attention.

PR #4374, opened 2019-12-13
adjust to current supported php versions

BlueTeck:
- removed PHP 5.6 (end of life since 1 year)
- added PHP 7.4 support

https://www.php.net/supported-versions.php


app/Core/User/UserSession.php
```diff
@@ -223,7 +223,7 @@ public function isLogged()
     public function getFilters($projectID)
     {
         if (! session_exists('filters:'.$projectID)) {
-            return session_get('user')['filter'] ?: 'status:open';
+            return session_get('user') ? session_get('user')['filter'] ?: 'status:open' : 'status:open';
```
fguillot: Why this condition need to be changed?

BlueTeck: I the unit tests session_get('user') could be null and the test failed when trying to access ['filter'] of null.

https://wiki.php.net/rfc/notice-for-non-valid-array-container

tests/units/Model/UserModelTest.php
```diff
@@ -392,6 +392,6 @@ public function testDisablePrivateProjects()
         $this->assertTrue($userModel->disable(2));
 
         $project = $projectModel->getById(2);
```
fguillot: This test was incorrect before your changes. Since we want to verify if a personal project is disabled when a user is disabled, it should be `$projectModel->getById(1)`. And the assertion below should be the same as before.

composer.json
```diff
@@ -14,7 +14,7 @@
         "discard-changes": true
     },
     "require" : {
-        "php" : ">=5.6.0",
+        "php" : ">=7.2.0",
```
fguillot: Which feature requires PHP 7.2? The `??` operator seems to have been introduced in PHP 7.0?

BlueTeck:
I will take a look and respond after the holidays

BlueTeck:
Reminder: https://github.com/kanboard/kanboard/blob/master/app/check_setup.php#L3

tests/units/Model/UserModelTest.php
BlueTeck: There was also another mistake in this test, the ower_id was 0 every time, therefore the project was not disabled, because it was not owned by user 2. Should be fine now.

composer.json
BlueTeck: Because PHP 7.0 and 7.1 were also outdated and not supported anymore. Why use another old version and give up new features that might be used. https://www.php.net/supported-versions.php

Issue #4359, opened 2019-11-24
Weblinks ending in ".io" are interpreted as attachments if External Link is set to "Auto"


smacz42:
### Actual behaviour

In a task, go to add a link as an "External Link". The title is not populate on the next screen, and the link shows up as an attachment, but with the correct hyperlink to the website.

### Expected behaviour

Links ending in `.io` should be recognized as weblinks.

### Steps to reproduce

Task -> "Add external link" -> "External Link": 'https://heliumjk.github.io', "Link Type": 'Auto' -> Title: 'Demo' -> "Save"

A new external link will show up, but instead of being labeled as a "Web link", it will show as an "Attachment".

### Configuration

- Kanboard version: v1.2.11
- Database type and version: 10.4.7-MariaDB-1:10.4.7+maria~bionic
- PHP version:7.3.8
- OS: Linux 4.4.0-157-generic
- Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.99 Safari/537.36 Vivaldi/2.9.1705.41

SladeP:
So looking at this, I found the problem. 

The Attachment provider's matching function looks for matches to this regex `preg_match('/^https?:\/\/.*\.([^\/]+)$/', $this->userInput, $matches)` . If a match is found, it extracts the extension and compares to this list:

`html, htm, xhtml, php, jsp, do, action, asp, aspx, cgi`

If the extension is NOT on that list, the link is considered an attachment because it's "not a web page". The problem is that the match will also occur for top level domains such as .com, .net, .io, etc. when they don't have a path after the domain. (https://www.google.com vs https://www.google.com/)

One solution would be to add several [top level domains](https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains) to that exclusive extension list, however this gets out of hand when it comes to countries (how do you differentiate between a website hosting a photoshop file (.ps) and a website for Palestine (google.ps)?

But we could keep it as simple as adding .com, .net, .io, .edu, .gov, and .org to the list and nothing else. But this is an international software, so we would be ignoring other country's top level domains and those urls would have the same issue.

Another solution is to force the web link provider matching function to run before the attachment provider. This is done by changing the code below so that the WebLinkProvider is added last (it's a LIFO queue).

app/ServiceProvider/ExternalLinkProvider.php
```
public function register(Container $container)
    {
[...]

PR #4382, opened 2020-01-02
Added some top level domains to the AttachmentLinkProvider match function

SladeP:
Fix for #4359 

Added some top level domains to the match function of the AttachmentLinkProvider so that .com, .net, .edu, .org, .io, and .gov are marked as web links not attachments. This will prevent links like https://www.github.io from being marked as attachments instead of web links. See my comment in the issue this fixes for more details.


fguillot:
Adjusting the regex seems to be a more generic fix.

For example: `https://www.example.org` matches and the TLD is used as extension which is wrong. Only URLs that contains a path should be taken into consideration (ie `https://www.example.org/file.pdf`).

SladeP:
Closing this pull request to open a new one with the fixes suggested.

PR #4384, opened 2020-01-03
Added setting that makes possible any new LDAP user to be Manager by default

JayBeeDe:
In big organizations, it is sometimes not possible to change the existing LDAP structure (OU, groups, etc.). It is not possible to make it satisfying the existing LDAP_GROUP_MANAGER_DN setting in order to get any new LDAP user to be Manager at its creation.
Adding the LDAP_USER_DEFAULT_ROLE_MANAGER boolean setting makes it possible to get any new user to be a Manager when set to true.
Of course this setting is disabled by default.


app/Core/Ldap/User.php
```diff
@@ -105,6 +105,17 @@ protected function getGroups(Entry $entry, $username)
         return $groupIds;
     }
 
+    /**
+     * Return true to set new user with role as Manager
+     *
+     * @access protected
+     * @return boolean
+     */
+    protected function userDefaultRole()
```
fguillot: It's probably more explicit to call this function something like `getDefaultUserRole()` and move the new condition here.

```php
protected function getDefaultUserRole()
{
    if (LDAP_USER_DEFAULT_ROLE_MANAGER) {
        return Role::APP_MANAGER;
    }
    return Role::APP_USER;
}
```

app/constants.php
```diff
@@ -81,6 +81,7 @@
 defined('LDAP_USER_ATTRIBUTE_PHOTO') or define('LDAP_USER_ATTRIBUTE_PHOTO', '');
 defined('LDAP_USER_ATTRIBUTE_LANGUAGE') or define('LDAP_USER_ATTRIBUTE_LANGUAGE', '');
 defined('LDAP_USER_CREATION') or define('LDAP_USER_CREATION', true);
+define('LDAP_USER_DEFAULT_ROLE_MANAGER') or define('LDAP_USER_DEFAULT_ROLE_MANAGER', false);
```
fguillot: There is a mistake here, the first part should be `defined`. You can see this error from the unit tests: `Warning: define() expects at least 2 parameters, 1 given in /home/travis/build/kanboard/kanboard/app/constants.php on line 84`

JayBeeDe: Yes

app/Core/Ldap/User.php
JayBeeDe: Hi, thank you for your quick reply. Yes do it!

Britaliope:
I think a `LDAP_USER_DEFAULT_ROLE` variable containing either `app-user`, `app-manager` or `app-admin` would be more flexible and future-proof (in case a new role is added).

Anyway @JayBeeDe do you think you will have time to fix the tests soon ? I would also be intrested in this feature/

JayBeeDe:
> I think a `LDAP_USER_DEFAULT_ROLE` variable containing either `app-user`, `app-manager` or `app-admin` would be more flexible and future-proof (in case a new role is added).
> 
> Anyway @JayBeeDe do you think you will have time to fix the tests soon ? I would also be intrested in this feature/

Yes I have/want to solve this. My changes are working fine for our customer. But the unit tests failed. @Britaliope I will have to think about it.

JayBeeDe:
> I think a `LDAP_USER_DEFAULT_ROLE` variable containing either `app-user`, `app-manager` or `app-admin` would be more flexible and future-proof (in case a new role is added).
> 
> Anyway @JayBeeDe do you think you will have time to fix the tests soon ? I would also be intrested in this feature/

@Britaliope Why would you need any new user to be Admin ? Give me an example with a context

JayBeeDe:
> The unit tests needs to be adjusted according to your changes. The build is currently marked as failed.

Hi,
@fguillot Sorry for the delay. I was waiting for our customer... I have changed (not removed) a test. All tests are passed now!

app/Core/Ldap/User.php
```diff
@@ -116,12 +127,18 @@ protected function getGroups(Entry $entry, $username)
...
     {
-        if (! $this->hasGroupsConfigured()) {
-            return null;
-        }
 
         // Init with smallest role
         $role = Role::APP_USER ;
+        
+        if (! $this->hasGroupsConfigured()) {
+            if ($this->userDefaultRole()) {
```
fguillot: This function is not really useful anymore. You can just use the constant here: `if (LDAP_USER_DEFAULT_ROLE_MANAGER) {...`

Issue #4119, opened 2019-01-09
Nginx reverse proxing on sub location not working with kanboard container


nicola-zanardi:
This issue is:

- [x] Bug report
- [ ] Feature request
- [ ] Improvement

### Actual behaviour

When configuring nginx to connect to the exposed port of the kanboard container, the connection fails as I cannot set a base url in kanboard. I tried setting the application URL in the settings, but it does not seem to help.

As a result, when I connect to `https://website.com/kanboard` I get redirected to `https://website.com/login` , and this does not work. If I force `https://website.com/kanboard/login`, I get a web page without CSS and that does not proceed further.

### Expected behaviour

I am able to set a base url to allow for reverse proxing, or I can configure my proxy server in a way that works as expected

### Configuration

- Kanboard version: 1.2.7
- Database type and version: as in the docker container
- PHP version: as in the docker container
- OS: Debian 9
- Browser: Firefox

My nginx configuration is now:

```
# Kanboard
location ~ ^/kanboard {
    proxy_pass http://127.0.0.1:99;
    proxy_set_header    Host            $host;
    proxy_set_header    X-Real-IP       $remote_addr;
    proxy_set_header    X-Forwarded-for $remote_addr;
}
```

but I already tried many many combinations without success.

kill0rz:
I have the same issue with apache webserver.

```
ProxyPass /kanboard/ http://kanboard:80/
ProxyPassReverse /kanboard/ http://kanboard:80/
```
I also get redirected to `/login` instead of `/kanboard/login`. Is there still no option to configure the URL in the docker version?

smacz42:
FWIW, I ran into the same issue - it's the same for a lot of docker apps. The fix is to simply symlink the directory to a similarly-named subdirectory of itself:

https://gitlab.com/smacz/role-compositional/blob/master/tasks/kanboard.yml#L64-66

It's a hack, but it works.

TheAssassin:
I also struggle with setting up kanboard on a "subdir". I've tried @smacz42's workaround, but that doesn't quite do the job. My Docker file:

```Dockerfile
FROM kanboard/kanboard:latest
RUN bash -xc "ln -sT /var/www/app/ /var/www/app/kanboard"
```

Then I changed my `proxy_pass` directive to point to the subdir. Unfortunately it doesn't help, when I request `/kanboard/`, I am now redirected to `/kanboard/login` first before being redirected to `/login` (before, that intermediate step was missing).

[...]

Issue #3534, opened 2017-11-05
URLs not working when behind nginx reverse proxy


XelaRellum:
We are using the docker image from https://hub.docker.com. On our docker server we have an nginx running that dispatches the URLs to the respective application (e.g. https://oursite.com/kanboard/ gets forwarded to http://localhost:4000). 

The login page shows up but all the assets link to an absolute path without /kanboard/ prefix, e.g. /assets/...

Our proxy rule for this site:
```
location /kanboard/ {
    proxy_pass http://127.0.0.1:8004/;
    proxy_redirect / /kanboard/;
  }
```

We have other applications running with the same set of rules (but different location). It seems as kanboard doesnt get the correct GET / LOCATION header.

- Kanboard version: current hub.docker.com image
- Database type and version: sqlite
- PHP version: 7
- OS: Linux
- Browser: Firefox

fguillot:
Did you try to define the application URL in settings?

XelaRellum:
Well was hard to get to the settings with the reverse proxy not working. Got around this by creating an SSH tunnel from my localhost to the docker port.
I changed the application url to the form "https://example.com/kanboard/" but the result was the same. For example the login screen still had links to the absolute "/assets" (without URL and without the application URL as prefix)

nealey:
I just got bit by this. Currently trying various approaches to build a new docker image. Would very much appreciate a configuration setting that says "this is my base path". This is a pretty standard option for web-based apps that allows reverse proxying.

bugwatcher:
This issue has been closed automatically by a bot for inactivity. For more information, read the [FAQ](https://docs.kanboard.org/en/1.2.6/developer_guide/faq.html).

PR #4395, opened 2020-01-28
Updates UrlHelper->dir() to use KANBOARD_URL if specified

freespace:
This resolves various redirection issues when hosted in a subdir, e.g.
https://example.com/kanboard/. See for example #4119 and #3534

### DIFF
--- a/app/Helper/UrlHelper.php
+++ b/app/Helper/UrlHelper.php
@@ -171,9 +171,13 @@ class UrlHelper extends Base
     public function dir()
     {
         if ($this->directory === '' && $this->request->getMethod() !== '') {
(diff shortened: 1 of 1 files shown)


fguillot:
The unit tests needs to be adjusted as well. Look at the logs: https://travis-ci.org/kanboard/kanboard/jobs/643141960?utm_medium=notification&utm_source=github_status

freespace:
@fguillot when you say "unit tests needs to be adjusted as well" do you mean I should add unit tests for this change?

fguillot:
Adding a test case and updating the docs would be nice.

Issue #3677, opened 2018-01-22
Setting for Task limit across swimlanes


n1trux:
Since v1.1.1 this has changed:

>Task limit apply across all swimlanes

Sadly I was super happy with the previous behaviour as I use different swimlanes for different customers.

Is it possible to get the old behaviour back? If not, I'd like to request this as a feature: Checkbox for "task limits across all swimlanes" in the settings per project.

PR #4399, opened 2020-01-31
Allow task limits to be applied per-swimlane

andrenth:
This PR adds a configuration toggle to projects that makes task limits local to each swimline. I believe this was how Kanboard used to work and it has been requested before (e.g. #3677).

### DIFF
--- a/app/Controller/ProjectCreationController.php
+++ b/app/Controller/ProjectCreationController.php
@@ -97,6 +97,7 @@ class ProjectCreationController extends BaseController
             'name' => $values['name'],
             'is_private' => $values['is_private'],
             'identifier' => $values['identifier'],
+            'per_swimlane_task_limits' => $values['per_swimlane_task_limits'],
         );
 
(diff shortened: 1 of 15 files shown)


BlueTeck:
Please take a look at the failing unit tests.

https://travis-ci.org/kanboard/kanboard/builds/644582377?utm_source=github_status&utm_medium=notification

andrenth:
Hm I ran the tests locally and only got an unrelated error related to LDAP. I’ll have a look.

andrenth:
It seems I have committed some commented code by mistake. I’ll fix it once I get home.

andrenth:
WIth the two latest pushes, unit tests pass for sqlite and postgres. MySQL tests are running but taking a long time.

andrenth:
So, the MySQL test failed, but I also get the same failure on master; it seem unrelated:
```
There was 1 failure:

1) TaskLinkModelTest::testGeyById
--- Expected
+++ Actual
@@ @@
-'is a child of'
+'is a parent of'
```

PR #4400, opened 2020-02-04
Fixes issue #4359. Updating the regex for matching attachment external links and added unit tests

SladeP:
Fix for #4359 

Updated the regular expression so that https://github.io will count as a web link and https://github.io/test.jpeg will count as an attachment. Also added a few lines to a unit test to ensure those types of links behave appropriately.

See #4359 for more comments on this solution.


SladeP:
One of my unit tests was using the wrong assertion. Fixed and committed and re-running unit tests.

SladeP:
Looks like Travis failed on composer install timeout. I don't know how to restart it.

PR #4403, opened 2020-02-06
Make tags global from project settings

BlueTeck:
Function to allow administrators to convert project tag to global tag.

image

image


BlueTeck:
Sorry, now everything should be in this PR.

PR #4414, opened 2020-02-20
Per-project and per-swimlane task limits

andrenth:
This PR allows projects and swimlanes to be configured with task limits that apply to their whole scope (i.e. all active tasks in a project or swimlane, respectively), as opposed to the usual per-column task limits.

### DIFF
--- a/app/Controller/BaseController.php
+++ b/app/Controller/BaseController.php
@@ -127,7 +127,7 @@ abstract class BaseController extends Base
     protected function getProject($project_id = 0)
     {
         $project_id = $this->request->getIntegerParam('project_id', $project_id);
-        $project = $this->projectModel->getByIdWithOwner($project_id);
+        $project = $this->projectModel->getByIdWithOwnerAndTaskCount($project_id);
 
         if (empty($project)) {
             throw new PageNotFoundException();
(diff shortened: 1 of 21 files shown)


andrenth:
Sorry for that. I've modified the tests to catch this, and fixed the query to work on Postgres.

fguillot:
I found another SQL issue with Postgres. When creating a new project, this error is raised: 

```
SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for integer: ""
```

The corresponding SQL is `INSERT INTO "projects" ("name", "is_private", "identifier", "per_swimlane_task_limits", "task_limit", "token", "last_modified", "owner_id") VALUES (:name, :is_private, :identifier, :per_swimlane_task_limits, :task_limit, :token, :last_modified, :owner_id)`

andrenth:
The latest commit adds `task_limit` to the list of numeric fields, and adds a test to catch this error.

fguillot:
Looks like there is a small glitch in the ui when creating a project with take limit:

image

The 0 is missing.

andrenth:
I hadn't realized the open task count was returned as NULL when the project has no tasks. I've changed the template to use `intval` on it.

PR #4445, opened 2020-03-23
enable or disable global tags

BlueTeck:
New option to enable or disable global tags per project

image


app/Schema/Mysql.php
```diff
@@ -8,7 +8,12 @@
 use Kanboard\Core\Security\Token;
 use Kanboard\Core\Security\Role;
 
-const VERSION = 136;
+const VERSION = 137;
+
+function version_137(PDO $pdo)
+{
+    $pdo->exec('ALTER TABLE `projects` ADD COLUMN `enable_global_tags` INT DEFAULT 1 NOT NULL');
```
rautamiekka: I dunno about PgSQL (I do know SQLite won't care what integer datatype you add) but I do know MySQL appreciates and recommends having as small integer datatype as possible which still fits the intended values, which means an INT for `0` and `1` values is total overkill and unnecessary performance **and** storage hit.

BlueTeck: thanks, fixed

BlueTeck:
I've created a new PR and changed some things based on your comments. #4454

PR #4446, opened 2020-03-24
Show group-memberships in user-Summary

manne65-hd:
I missed a quick way to see the group-memberships of a given user, so I added this.

I hope my approach complies with kanboard's coding guidelines.

PR #4449, opened 2020-03-28
Show group membership(s) in user-summary and user-list

manne65-hd:
I missed a quick way to see the group-membership(s) of a given user, so I added this feature.

### Screenshots
**User-summary-page**
_There's a new line beneath the user role-information, that displays all group-memberships for the current user_
UserInformation

**User-list**
_In the userlist, you can see the group-membership(s) of each user. If a user has no group-memberships, no additional info is displayed_
UserListNoLimit

### Information
This is my second attempt at that feature, while [the first version (PR 4446)](https://github.com/kanboard/kanboard/pull/4446) had 2 "issues"
* Displaying group-memberships in user-list was not implemented
* Travis-Checks had gotten stuck somehow
*  Therefore: PR 4446 closed

### DIFF
--- a/app/Helper/UserHelper.php
+++ b/app/Helper/UserHelper.php
@@ -110,7 +110,7 @@ class UserHelper extends Base
     }
 
     /**
-     * Get group names(as a comma-separted list) for a given user
+     * Get group names(as a comma-separated list) for a given user
      *
      * @access public
      * @param  integer   $user_id   User id
--- a/app/Helper/UserHelper.php
+++ b/app/Helper/UserHelper.php
@@ -110,6 +110,18 @@ class UserHelper extends Base
     }
 
     /**
+     * Get group names(as a comma-separted list) for a given user
+     *
+     * @access public
+     * @param  integer   $user_id   User id
+     * @return string
+     */
+    public function getGroupNames($user_id)
+    {
+        return implode(', ', array_column($this->groupMemberModel->getGroups($user_id), 'name'));
+    }
+
+    /**
      * Check application access
      *
      * @param  string  $controller
--- a/app/Locale/bs_BA/translations.php
+++ b/app/Locale/bs_BA/translations.php
@@ -838,6 +838,7 @@ return array(
     'Remove group' => 'Ukloni grupu',
     'Group removed successfully.' => 'Grupa uspješno uklonjena.',
     'Unable to remove this group.' => 'Nemoguće ukloniti grupu.',
+    // 'Group membership(s):' => ':',
     'Project Permissions' => 'Prava na projektu',
     'Manager' => 'Menadžer',
     'Project Manager' => 'Menadžer projekta',
--- a/app/Locale/ca_ES/translations.php
+++ b/app/Locale/ca_ES/translations.php
@@ -838,6 +838,7 @@ return array(
     'Remove group' => 'Elimina el grup',
     'Group removed successfully.' => 'Grup eliminat correctament.',
     'Unable to remove this group.' => 'No es pot eliminar aquest grup.',
+    // 'Group membership(s):' => ':',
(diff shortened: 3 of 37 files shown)


manne65-hd:
### Request for comments
As I am well aware that _"Kanboard focuses on simplicity and minimalism"_, I am uncertain, whether adding 2 more options to this new feature would be welcome or would rather break that KISS-principle?

**Additional option 1: > Switch OFF display of group-membership(s) in user-list**
I have already prepared an option, to disable the display of group-membership(s) in the user-list. Like this, the group-membership(s) would only be displayed on the user-summary-page.

**Additional option 2: > Limit number of group-membership(s) displayed in user-list**
While this sounds stupid in the first place, it is intended at Kanboard-installations, that might have many groups and at least some users, that are members in many groups at once as this could end up in the user-list to look "ugly".
If a limit is set, the number of group-memberships displayed in the user-list will be limited to that number. 
The "magic" comes in, because there will be a hint added to that list, telling the user to hover his mouse over the group-icon, to reveal the complete list. _(See screenshot below)_
UserListLimited

**How is this done?**
I implemented this by adding to new CONSTANTS to _config.php_ like so:
```
// Enable or disable displaying group-memberships in userlist (true by default)
define('SHOW_GROUP_MEMBERSHIPS_IN_USERLIST', true);

// Limit number of groups to display in userlist (The full list of group-memberships is always shown, when hovering the mouse over the group-icon of a given user!)
// If set to 0 ALL group-memberships will be listed (0 by default = do not limit)
define('SHOW_GROUP_MEMBERSHIPS_IN_USERLIST_WITH_LIMIT', 0);

```
And finally my code is also prepared to handle these 2 CONSTANTS to be "optional", meaning that they would not necessarily have to be added to the config.default.php and what's even more important ... it wouldn't break things in existing installations, where users could forget to add the 2 CONSTANTS to their existing config.php after an update.

Let me know what you think.

Cheers,
Manfred

manne65-hd:
As I am currently trying to learn and understand how to code the "kanboard"-style, I have prepared another way of controlling the 2 new options from within the ADMIN-settings-panel.

Looking forward to hear which version you would prefer to merge into the codebase _(if at all)_

**Here's a screenshot of my current attempt to control via panel**
SettingsWithDefault

fguillot:
I like the idea of showing the popup when the user have lot of groups. It doesn't need to be configurable.

manne65-hd:
> I like the idea of showing the popup when the user have lot of groups. It doesn't need to be configurable.

Hey Fred,
thanks for pulling this into the codebase :-)
If you like the idea of reducing the full list and only displaying that in the popup, we **must** decide how exactly this is supposed to be controlled, because the current code will always display the full list of groups for every user in the user-list!

**Here's my proposal how to do that:**
- Preconfigure a fixed limit of 7 groups into the code
- Enable the option of overriding/disabling that fixed limit, by adding an **optional** CONSTANT into config.php 
```
// Limit number of groups to display in userlist (The full list of group-memberships is always shown, when hovering the mouse over the group-icon of a given user!)
// If set to 0 ALL group-memberships will be listed (7 by default)
define('SHOW_GROUP_MEMBERSHIPS_IN_USERLIST_WITH_LIMIT', 7);
```

I would also propose to stick with my current idea of disabling the whole thing by also using another **optional** CONSTANT in config.php like so:
```
// Enable or disable displaying group-memberships in userlist (true by default)
define('SHOW_GROUP_MEMBERSHIPS_IN_USERLIST', true);
```

**Advantages**
- As both CONSTANTS are optional, everything will also work if these are not present in config.php which is good for existing installations, where users might not add the 2 new CONSTANTS to their existing config.php
- If an admin of a Kanboard-installation wants to control the 2 new features, he/she is covered!

Looking forward to hear what you think.

PR #4453, opened 2020-04-03
Show group memberships v3

manne65-hd:
Referring to [your comment](https://github.com/kanboard/kanboard/pull/4449#issuecomment-608219904) on the merged PR# 4449 ...

This should also be merged, to enable the feature of the reduced list of groups!

### DIFF
--- a/app/Helper/UserHelper.php
+++ b/app/Helper/UserHelper.php
@@ -110,6 +110,18 @@ class UserHelper extends Base
     }
 
     /**
+     * Get group names(as a comma-separated list) for a given user
+     *
+     * @access public
+     * @param  integer   $user_id   User id
+     * @return string
+     */
+    public function getGroupNames($user_id)
+    {
+        return implode(', ', array_column($this->groupMemberModel->getGroups($user_id), 'name'));
+    }
+
+    /**
      * Check application access
      *
      * @param  string  $controller
--- a/app/Helper/UserHelper.php
+++ b/app/Helper/UserHelper.php
@@ -110,45 +110,33 @@ class UserHelper extends Base
     }
 
     /**
-     * Check if group-memberships should be displayed in user-list for a given user
-     * User must be a member of at least one group && CONSTANT(from config.php) to diplay group-memberships in userlist must be true!
+     * Get group names for a given user and return an associative array:
+     *  ['full_list'] = a comma-separated list of all group-memberships
+     *  ['limited_list'] = a comma-separated list limited to N groups depending on value of SHOW_GROUP_MEMBERSHIPS_IN_USERLIST_WITH_LIMIT
+     *  ['has_groups'] = boolean TRUE if user is member of at least one group ... else FALSE
      *
      * @access public
      * @param  integer   $user_id   User id
-     * @return boolean
+     * @return array
      */
-    public function getDisplayGroupNamesInUserList($user_id)
+    public function getUsersGroupNames($user_id)
     {
-        return (count($this->groupMemberModel->getGroups($user_id)) > 0 && SHOW_GROUP_MEMBERSHIPS_IN_USERLIST);
-    }
+        $groups_list = array_column($this->groupMemberModel->getGroups($user_id), 'name');
+        $full_list = implode(', ', $groups_list);
 
-    /**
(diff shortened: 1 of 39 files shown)


manne65-hd:
> My only concern is the number of SQL queries generated. The same function `$this->groupMemberModel->getGroups()` is called many times for a single user. 

You are right ... I didn't think about that 

> With few changes, you should be able to call this function only once per user.

Changed my code to reduce the calls of `$this->groupMemberModel->getGroups()` from 3 to 1 per user.
In addition I also changed it to avoid any additional SQL-queries, if the new config-CONSTANT `SHOW_GROUP_MEMBERSHIPS_IN_USERLIST` is set to false

But now there's MERGE conflicts because another merge has also changed the locales. I am still pretty new with git and github, so I don't really know how to resolve that.
Can you help?

app/Template/user_list/user_details.php
```diff
@@ -11,7 +11,10 @@
         <span><a href="mailto:<?= $this->text->e($user['email']) ?>"><?= $this->text->e($user['email']) ?></a></span>
     <?php endif ?>
 
-    <?php if (! empty($this->user->getGroupNames($user['id'])) ): ?>
-        <span><i class="fa fa-fw fa-group aria-hidden="true"></i> <?= $this->user->getGroupNames($user['id']) ?></span>
+    <?php if ( SHOW_GROUP_MEMBERSHIPS_IN_USERLIST ): ?>
+        <?php $groups_list_tooltip = t('%s is a member of the following group(s):', $user['name']) . '&#10;' . $this->user->getUsersGroupNames($user['id'])['full_list']; ?>
+        <?php if ($this->user->getUsersGroupNames($user['id'])['has_groups']): ?>
+            <span><i class="fa fa-fw fa-group aria-hidden="true" title="<?= $groups_list_tooltip ?>"></i> <?= $this->user->getUsersGroupNames($user['id'])['limited_list'] ?></span>
```
fguillot, line 17: When you are doing this `$this->user->getUsersGroupNames($user['id'])['limited_list']`, PHP will call your function first and then return the result. That means you are still executing 3 queries. You should store the result in a variable to avoid this.

For example:

```php
<?php $userGroups = $this->user->getUsersGroupNames($user['id']); ?>
```

And then you just use your array: `$userGroups['limited_list']`.

app/Helper/UserHelper.php
```diff
@@ -110,15 +110,33 @@ public function getRoleName($role = '')
...
+        $limited_list = ( $groups_list == $limited_list ) ? implode(', ', $limited_list) : implode(', ', $limited_list) . ' ( >> ' . t('hover mouse over group-icon, to show all group-memberships') . ' )';
+
+        $has_groups = (count($groups_list)) ? true : false;
+
+        $UsersGroupNames = array(
+            'full_list' => $full_list,
+            'limited_list' => $limited_list,
+            'has_groups' => $has_groups
+        );
+        return $UsersGroupNames;
```
fguillot, line 139: Small improvement here: just return the array directly. The variable `$UsersGroupNames` is not necessary.

app/Template/user_list/user_details.php
manne65-hd: I am a little embarrassed, because I was so focused on reducing the 3 methods in the helper to just 1, that I forgot about reducing the actual method-calls :-/

I guess that's what code-reviews are good for 👍 

Thanks for your help ... hope I solved all issues now.

manne65-hd:
Just out of curiosity ... I have 2 more questions:
- **Always append new translation-strings to the end of the translations.php-file(s) ?**
  - It looks like other contributors add newly introduced language-strings to end of the translations.php-file(s) ...
  - Would that have solved my MERGE-conflicts automatically ?
  - If not .. I guess it wold at least make it easier to solve them manually ?
- **Why no unit-tests for MySQL ?**
  - I am still new to all this stuff around github / unit-tests / etc. , but it seems to me, that unit-tests for kanboard-pull-requests are only run with SQlite and PostgreSQL. Why is that ?

Looking forward to your answers,
Manfred.

P.S.:
BTW thanks for a great pice of software 👍

PR #4460, opened 2020-04-05
Final fixes / german translation for user-groups-listing

manne65-hd:
I like your new solution to tell the user, that there are more groups for a given user as well as the tooltip wrapped inside the whole <span>
But I assume that you intended  this to read as "‑ **shown** / **all** group(s) shown" ?

I also re-introduced the feature, that this additional information is only displayed, if the list for a user is limited at all, because at least that's the way I intended it to work :-)


fguillot:
👍

PR #4466, opened 2020-04-10
Per project tasks numeration

mrozigor:
This is only proposal of feature and solution - still WiP (in terms of incomplete code and formatting).

Do you think that this will be useful? If so, then is this a correct way? Are those all entries in system for task creation? Or subtasks also are saved in same table and use common ID with tasks?

If you think that this is a good idea, then I'll be thankful for any answers for above questions and then will proceed with coding rest of this feature.


fguillot:
There is something wrong with the vendor folder (see https://docs.kanboard.org/en/latest/developer_guide/faq.html#why-php-vendor-directory-is-committed-into-the-source-tree). It's hard to review the changes.

mrozigor:
Reverted changes with `vendor` directory.

mrozigor:
I know - I just want to know if change is going in the right direction. If so, then I'll have to add new task IDs in different places (ID that consists of project code and project_task_id - like in Jira).

fguillot:
All these operations should probably be done in a transaction. I'm sorry but that seems to be too experimental to be merged.

pehua:
> I know - I just want to know if change is going in the right direction. If so, then I'll have to add new task IDs in different places (ID that consists of project code and project_task_id - like in Jira).

Hi, any update about this feature? Thanks

mrozigor:
Nope, abandoned PR ;)

PR #4484, opened 2020-04-24
Also copy over subtask assignee on duplicate

### LINKED ISSUE #4469: Recurrent task does not copy over subtask assignees
### Actual behaviour

When a recurrent task is created it copies over the subtasks but fails to copy over subtask assignees

### Expected behaviour

Subtask assignees should also be copied

### Steps to reproduce

Just create a recurrent task with a subtask and assign it to someone, then force the recurrence to occur and look at the copied task.

### Configuration

- Kanboard version: dev-version from github on 2020-04-07
- Database type and version: sqlite3
- PHP version: 7.3
- OS: Ubuntu 18.4 LTS
- Browser: firefox/chrome/...

liedekef:
This resolves #4469

### DIFF
--- a/tests/units/Model/SubtaskModelTest.php
+++ b/tests/units/Model/SubtaskModelTest.php
@@ -152,6 +152,9 @@ class SubtaskModelTest extends Base
         $this->assertEquals(0, $subtasks[0]['status']);
         $this->assertEquals(0, $subtasks[1]['status']);
 
+        $this->assertEquals(0, $subtasks[0]['user_id']);
+        $this->assertEquals(1, $subtasks[1]['user_id']);
+        
         $this->assertEquals(1, $subtasks[0]['position']);
         $this->assertEquals(2, $subtasks[1]['position']);
     }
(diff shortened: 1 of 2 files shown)


BlueTeck:
The unit test is failing:

https://github.com/kanboard/kanboard/blob/master/tests/units/Model/SubtaskModelTest.php#L156

liedekef:
Yes, it seems the unit test is checking that user_id is empty. So should I change my pull request to adapt the unit test too?

tests/units/Model/SubtaskModelTest.php
```diff
@@ -152,9 +152,6 @@ public function testDuplicate()
         $this->assertEquals(0, $subtasks[0]['status']);
         $this->assertEquals(0, $subtasks[1]['status']);
 
-        $this->assertEquals(0, $subtasks[0]['user_id']);
-        $this->assertEquals(0, $subtasks[1]['user_id']);
-
```
fguillot, line 157: The unit test should validate that the assignee is correctly copied.

liedekef: Agreed, I'll adapt the unit test again

PR #4488, opened 2020-04-27
Make tooltip events bubble

### LINKED ISSUE #4483: Popups on hover not working after moving a task between columns
### Actual behaviour

When you have a task with subtasks, you can hover over the icon to see the subtasks in the kanboard view. However, after moving a task, this hover no longer pops up the subtasks (for the move task or any other task). Refreshing the window is needed to make that work again.

### Expected behaviour

Hovering should always show the expected popup.

### Logs

Unfortunately no logs, not even in the browser console. So I assume some javascript is blocking it based on a security policy after moving a task.

### Configuration

- Kanboard version: latest dev version
- Database type and version: mariadb
- PHP version: 7.2
- OS: Ubuntu 18.04 LTS
- Browser: Firefox 75.0

liedekef:
Make sure tooltips work even after page content change (as when a task moves from one column to another). This fixes #4483


fguillot:
I just noticed that you also need to include the minified Javascript in your PR (run `./cli js`).

I'm not able to reproduce the issue in the first place. What do you do exactly?

liedekef:
Well, if you move a task on the kanboard (from one column to another), it triggers a ajax call. The reply of that call is html that replaces the kanboard content completely. But as a result of that, all event tooltip listeners are no longer valid and so tooltips no longer work (like when hovering over the subtask icon, task link icon etc ...). This is reproducable in Firefox and Chrome.

liedekef:
Well,I delved further into the code and KB.tooltip function gets called after drag and drop, so it should be working.
So I went back to testing and it seems Bigboard is causing the problem. Although my fix seems to working even when Bigboard is installed :-)

PR #4491, opened 2020-04-29
Keep newlines in markdown

liedekef:
Doesn't break existing functionality and improves readability


fguillot:
The preview button above the texarea still show the content without break line.

liedekef:
> The preview button above the texarea still show the content without break line.

I confirm that, but I believe that to be a bug with ParseDown ... maybe it is already fixed there, that I don't know. I'll check their github and maybe open a bug report for that there.

fguillot:
> I think it would be wiser to replace that with an ajax call and use the php markdown parser lib to render the text, so it is identical everywhere (and be less dependand on another lib).

👍 Good idea

PR #4493, opened 2020-04-30
 Optimize for mobile

plague69:
Snapping horizontal scrolling for swimlanes
Column spacing for finger placement for sliding
Minify header into
Views switcher into buttons instead of taking up whole width


plague69:
image

fguillot:
The view switcher buttons should have some space between them, no?

btw, you should also include the minified css in your PR: https://docs.kanboard.org/en/latest/developer_guide/assets.html

plague69:
image
- Buttons share same spacing as columns
- Minified compiled

fguillot:
The button labels overflows when using a different language. Here an example in French:

image

plague69:
image
Best solution I can find at the moment, just wrap the word but will be multi line

k00ni:
Thank you for doing this, looking forward to a merge.

nekohayo:
> Best solution I can find at the moment, just wrap the word but will be multi line

@plague69 in my humble opinion (I'm not a dev/maintainer, just a user): when space is at a premium, I believe it would be better to reduce the padding and hide the icons for those three mode switcher buttons at the top (and _maybe_ reduce the font size by 10% or so, if that's still not sufficient). Label-only buttons are always comprehensible by users, icon-only ones are not, and I do not consider those icons to be strictly required here anyway.

My proposed solution might not work for _every_ language, but it probably will for a good proportion of them.

plague69:
> My proposed solution might not work for _every_ language, but it probably will for a good proportion of them.
It's a good idea, I'll see if I have time to change it up after I merge all the new commits into my old branch and test for conflicts. The original changes have been sitting open for nearly 10 months so I guess it will only ever be used as a reference and never be merged into main.
> I don't quite see why you separated those three buttons instead of keeping them visually linked together, however...
In regards to this one. I agree screen real-estate is valuable. I was thinking some people might still be using smaller devices (like the iPhone 4 sized screens). Disconnected/spaced creates more of a safe zone so people won't mistakenly tap. Removing icons is doable as well. I'll see which language has the longest name and see if it fits single line without the icon. French still seems to be close to limit

fguillot:
Any particular reason why the login form is centered vertically?

image

I noticed some misalignments here and there compared to the actual version. For example, the closing icon in modal box:

image

xam-ps:
Would be awesome to see this beeing merged in one of the next versions. Right now using KB on mobile is quite a pain 😅

PR #4496, opened 2020-05-02
create tooltip listeners at dom.ready only

liedekef:
creating tooltip listeners at dom.ready only now that they are generic to the body, otherwise listeners will be added on every ajax action


assets/js/src/BoardDragAndDrop.js
```diff
@@ -103,7 +103,6 @@ Kanboard.BoardDragAndDrop.prototype.refresh = function(data) {
     this.app.hideLoadingIcon();
     this.executeListeners();
     this.dragAndDrop();
-    KB.tooltip();
```
fguillot, line 106: Only this one needs to be removed?? the `tooltip()` function is already executed when the DOM is ready.

liedekef: No, I removed the function name and changed it so the whole tooltip section gets executed at dom ready. So no need for a separate function call/name anymore.
If I kept the function name then your suggestion would be sufficient too, but I don't like unneeded function names :-)

PR #4498, opened 2020-05-04
Add subtask events to ProjectModificationDateSubscriber

liedekef:
This allows also subtask updates to be reflected on the dashboard (the % info) when a subtask is completed.

### DIFF
--- a/app/Subscriber/ProjectModificationDateSubscriber.php
+++ b/app/Subscriber/ProjectModificationDateSubscriber.php
@@ -4,6 +4,7 @@ namespace Kanboard\Subscriber;
 
 use Kanboard\Event\GenericEvent;
 use Kanboard\Model\TaskModel;
+use Kanboard\Model\SubtaskModel;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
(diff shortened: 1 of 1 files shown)


fguillot:
I don't see the percentage updated when changing the status of subtasks. How can I test this exactly?

liedekef:
Well, indeed nu setting the subtask status to completed and then wait for the next Ajax board refresh. I'm using this in production since last week and it works fine.

fguillot:
The Ajax request contains only the element `<div class="task-list-subtasks">`. This is why I don't see the percentage updated.

image

Are you using this in addition to some plugin?

liedekef:
I just tried this on a virgin setup and it works ok. But according to your screenshot you're testing this in the controller=DashboardController&action=show view (the "main" dashboard view), while I tested this in the per-project Board view (e.g. controller=BoardViewController&action=show&project_id=1 ).
It seems the main dashboard doesn't have a planned ajax call that refreshes every X seconds, so I consider that normal. This change just adds to the per-project Board ajax calls.

liedekef:
Hi, would it be possible to merge this?

fguillot:
I can see it now.

PR #4500, opened 2020-05-05
Markdown tooltip fix

liedekef:
The p-margin was being overwritten for all p-tags in the tooltip, while it should only do that for the last child


fguillot:
I don't see the minified file? :)

liedekef:
> I don't see the minified file? :)

It seems github refuses a direct upload of a minified css ... I'll need to do this the old fashioned local way :-)

PR #4509, opened 2020-05-14
Create SECURITY.

leryan06:


fguillot:
What is this? The content is copy pasted from a template without being specific to Kanboard.

PR #4513, opened 2020-05-19
Allow use of the user's DN as the group filter substitution

mildis:
* create the `LDAP_GROUP_USER_ATTRIBUTE` property, defaulting to `username`  
(other possible value : `dn`)


ptoulouse:
It seems that /docker/etc/php7/php-fpm.d/env.conf was not updated to add LDAP_GROUP_USER_ATTRIBUTE.

PR #4546, opened 2020-06-18
Added standard notification footer to comment-mail template

kinger-de:
A few users with us wished to have the standard notification/footer in the comment emails. For this I hand over the task to the template in the controller.

### DIFF
--- a/app/Controller/CommentMailController.php
+++ b/app/Controller/CommentMailController.php
@@ -49,7 +49,7 @@ class CommentMailController extends BaseController
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/Controller/CommentMailController.php
```diff
@@ -49,9 +49,9 @@ public function save()
         }
     }
 
-    protected function sendByEmail(array $values)
+    protected function sendByEmail(array $values, $task => null)
```
fguillot, line 52: This syntax is not valid in PHP.

PR #4557, opened 2020-07-02
new config parameter SESSION_HANDLER

matthiasstraka:
I added a setting SESSION_HANDLER that allows to disable database-backed session handling (e.g. set SESSION_HANDLER to "php" or any value that is not "db" for now). The reason for this change is that I like my server to limit writes to the database and instead use php session files in a tmpfs store. With this change, I can navigate kanboard pages without any writes to the database.

### DIFF
--- a/app/Core/Session/SessionManager.php
+++ b/app/Core/Session/SessionManager.php
@@ -38,7 +38,9 @@ class SessionManager extends Base
      */
     public function open()
     {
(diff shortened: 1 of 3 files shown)


fguillot:
That would be appreciated if you can update the [docs](https://github.com/kanboard/documentation) as well.

PR #4561, opened 2020-07-06
Add option to configure SMTP HELO name

sxntxn:
Some SMTP setups check the HELO/EHLO name supplied by the connecting
client. Allow this to be configured explicitly.

(This is needed for the MTA my installation works with...)


app/Core/Mail/Transport/Smtp.php
```diff
@@ -23,6 +23,8 @@ protected function getTransport()
         $transport = Swift_SmtpTransport::newInstance(MAIL_SMTP_HOSTNAME, MAIL_SMTP_PORT);
         $transport->setUsername(MAIL_SMTP_USERNAME);
         $transport->setPassword(MAIL_SMTP_PASSWORD);
+	if (!is_null(MAIL_SMTP_HELO_NAME))
```
fguillot, line 26: The indentation seems inconsistent with the rest.

fguillot: And if you can use curly braces for your if statement..

fguillot:
It would be nice if you could add the new option in the documentation: https://github.com/kanboard/documentation

PR #4600, opened 2020-09-13
Fix : Prevent the original page from being modified by the opened link

mildis:


fguillot:
Only the first commit should be part of this pull-request.

mildis:
True.
It wasn't the same branches, I think I messed up here. I'll close this one and open a new only for the first commit.

PR #4601, opened 2020-09-13
fix,minor: replace <i> by <em>

mildis:


fguillot:
Why?

FontAwesome uses `<i>` tag. If it's about semantics, `<span>` is more appropriate according to their docs.

https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use

> Font Awesome is designed to be used with inline elements, and we recommend sticking with a consistent HTML element to reference them by in your project. We like the `<i>` tag for brevity and because most folks use `<em></em>` for emphasized/italicized semantic text these days. If that’s not your cup of tea, using a `<span>` is more semantically correct.

Issue #4070, opened 2018-11-14
title attribute is not accessible


dnz3d4c:
I am using Kanboard with a screen reader.
When using Kanboard as a screen reader, it is inconvenient to access the document with the cursor of the screen reader because of the title attribute.
In many places, the title attribute mentions poor accessibility. You can replace it with an alternative solution through the following resources:
1. https://a11yproject.com/posts/title-attributes/
2. https://developer.paciellogroup.com/blog/2010/11/using-the-html-title-attribute/
3. https://www.mediacurrent.com/blog/dont-rely-title-attribute-accessibility-seo/

Typically, the following elements of Kanboard are provided using the title attribute:
1. default filter link in the search area
2. Configure this project link

The title attribute may be used in more places. I hope this issue has been fixed and it will be easier to use Kanboard as a screen reader. Thank you.

bugwatcher:
Please, use the [forum](https://kanboard.discourse.group) for open discussions, suggestions and feature requests. **This bug tracker is reserved only for bugs**.

Kolossi:
The bug is that the title attribute is used when it should not be, which is an issue when using a screenreader - in the same way that there's always a subset of uses that lead to _any_ bug.

OP has given good resources for someone to address this. If "bugwatcher" is automated, maybe that's why it thought it was a discussion?

I'd agree with OP that this is a bug and wish I had the time to address it.

michaelvickersuk:
@dnz3d4c I've made a number of improvements to make the title attributes across Kanboard more accessible, hopefully this will improve things for screen reader users. It would be great to hear your feedback [when the new version gets released](https://github.com/kanboard/kanboard/releases).

michaelvickersuk:
These improvements have now been included as part of [release 1.2.16](https://github.com/kanboard/kanboard/releases/tag/v1.2.16)

PR #4617, opened 2020-10-03
Hidden accessible titles

michaelvickersuk:
For elements which have a title attribute and don't have a suitable element to which to add a `aria-label`, provide the title text hidden inline but still accessible to screen readers. Resolves #4070.

### DIFF
--- a/app/Helper/TaskHelper.php
+++ b/app/Helper/TaskHelper.php
@@ -238,7 +238,7 @@ class TaskHelper extends Base
     public function renderPriority($priority)
     {
         $html = '<span class="task-priority" title="'.t('Task priority').'">';
-        $html .= '<span class="ui-helper-hidden-accessible">'.t('Task priority').': </span>';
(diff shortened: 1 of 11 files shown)


app/Helper/TaskHelper.php
```diff
@@ -238,6 +238,7 @@ public function renderDueDateField(array $values, array $errors = array(), array
     public function renderPriority($priority)
     {
         $html = '<span class="task-priority" title="'.t('Task priority').'">';
+        $html .= '<span class="ui-helper-hidden-accessible">'.t('Task priority').': </span>';
```
fguillot, line 241: The colon should be part of the translated text because typographic rules are different between languages.

For example, in French, there is a non-breaking space before the colon: `Priorité de la tâche : P1`.

michaelvickersuk: Thanks for the info :slightly_smiling_face: I never knew that differed between languages.

I've removed the colons so we aren't introducing further translation terms and can simply make use of the existing title terms.

PR #4618, opened 2020-10-03
Correct table collapsed column titles

michaelvickersuk:
Correct the title attributes used when a board column is collapsed, so that it matches with the element's actual purpose. 

Also provide the title text hidden inline so it's accessible to screen readers, resolves #4070

# Screenshots

Element | Before | After
------------ | ------------- | -------------
Column header task count | image | image
Column label | image | image
Expand column button | image | image

### DIFF
--- a/app/Template/board/table_column.php
+++ b/app/Template/board/table_column.php
@@ -8,8 +8,8 @@
 
         <!-- column in collapsed mode -->
         <div class="board-column-collapsed">
-            <small class="board-column-header-task-count" title="<?= t('Show this column') ?>">
-                <span id="task-number-column-<?= $column['id'] ?>"><?= $column['nb_tasks'] ?></span>
+            <small class="board-column-header-task-count" title="<?= t('Task count') ?>">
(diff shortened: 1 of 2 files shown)


app/Template/board/table_column.php
```diff
@@ -8,8 +8,8 @@
 
         <!-- column in collapsed mode -->
         <div class="board-column-collapsed">
-            <small class="board-column-header-task-count" title="<?= t('Show this column') ?>">
-                <span id="task-number-column-<?= $column['id'] ?>"><?= $column['nb_tasks'] ?></span>
+            <small class="board-column-header-task-count" title="<?= t('Task count') ?>">
+                <span id="task-number-column-<?= $column['id'] ?>"><span class="ui-helper-hidden-accessible"><?= t('Task count') ?>: </span><?= $column['nb_tasks'] ?></span>
```
fguillot, line 12: Same comment as this one: https://github.com/kanboard/kanboard/pull/4617#discussion_r499268766

michaelvickersuk: Colons now removed so we can simply re-use the existing translation terms

PR #4619, opened 2020-10-03
Add ARIA label to form text editor without label

michaelvickersuk:
For form text editors that do not have a visual label element, use aria-label to provide an accessible string. The ARIA labels make use of existing translation terms therefore these will be available in all languages. Resolves #4070.


fguillot (commented):
The minified Javascript should be part of the pull-request. You can find the instructions on this page: https://docs.kanboard.org/en/latest/developer_guide/assets.html

fguillot:
Since you have sent lot of pull-requests, I will regenerate the Javascript after the merges, otherwise it will creates conflicts.

PR #4642, opened 2020-10-17
Fix tasks.swimlane_id foreign key for sqlite

### LINKED ISSUE #4635: Task moved to inexistent swimlane when providing an incorrect swimlane_id
### Actual behaviour

By mistake my API call was moving (`moveTaskPosition`) a task in column DONE, in `swimlane_id` 2 which no longer exists.

This doesn't trigger any error and the task get kind of lost.

I discovered the bug when I saw tasks count reporting a higher number of tasks than actually displayed, and some recent done tasks missing.

We can still find the the task in the overview though, and we see it gets moved to `""`.

image

The task can't be recovered from the board view:

image

### Expected behaviour

I understand this is a mistake on my end but the task should not get moved if an incorrect `swimlane_id` is provided.

### Configuration

    Application version: v1.2.16
    PHP version: 7.3.23
    PHP SAPI: fpm-fcgi
    HTTP Client: cURL
    OS version: Linux 5.1.20-200.fc29.x86_64
    Database driver: sqlite
    Database version: 3.32.1
    Browser: Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:80.0) Gecko/20100101 Firefox/80.0

bu4ak:
Closes #4635 

Add reference `tasks.swimlane_id` -> `swimlane.id` for sqlite. 
I found this is already implemented in mysql and postgres migrations.


app/Schema/Sqlite.php
```diff
@@ -8,7 +8,52 @@
...
+            score                INTEGER,
+            date_due             INTEGER,
+            category_id          INTEGER DEFAULT 0,
+            creator_id           INTEGER DEFAULT '0',
+            date_modification    INTEGER DEFAULT '0',
+            reference            TEXT    DEFAULT '',
+            date_started         INTEGER,
+            time_spent           NUMERIC DEFAULT 0,
+            time_estimated       NUMERIC DEFAULT 0,
+            swimlane_id          INTEGER REFERENCES swimlanes ON DELETE CASCADE,
```
fguillot, line 38: It would be better to make the foreign column explicit, for example: `REFERENCES swimlanes(id)` (same for `projects` and `columns` tables above).

bu4ak: done

Issue #4640, opened 2020-10-15
Role lost on LDAP login after upgrade to 1.2.16


rafacamargo123:
### Actual behaviour

After loggin in with LDAP the user loses its current setted roles (admin, manager) and is setted as USER.

### Expected behaviour

The previous behaviour on kanboard allowed users to have roles setted manually.

### Steps to reproduce

login with LDAP with an user that doesn't matches LDAP_GROUP_ADMIN_DN or LDAP_GROUP_MANAGER_DN

login with an admin user and set the previous user as ADMIN or MANAGER

login again with the previous user. His role is removed

### Configuration

- Versão da aplicação: v1.2.16
- Versão do PHP: 7.3.23
- PHP SAPI: fpm-fcgi
- Cliente HTTP: cURL
- Versão do sistema operacional: Linux 3.10.0-1062.4.1.el7.x86_64
- Driver do banco de dados: mysql
- Versão do banco de dados: 5.7.9-log
- Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.80 Safari/537.36

I believe this behaviour was introduced by e3e9cabd8bfba57eedd4e820bcd6c0ace34eb80c - #4384

Floyer007:
Same behaviour for me as well.

Only difference: I'm using sqlite as database-driver.

inful:
Same problem

simonjamain:
same

ChristianSwift:
楼上几位大哥都不读文档的吗？[点我](https://docs.kanboard.org/en/1.2.16/admin_guide/ldap_authentication.html#synchronization)

Floyer007:
How should this solve the problem?
Everything in the list will _not_ be synchronized.

I've only "username" in my list.

ChristianSwift:
> How should this solve the problem?
> Everything in the list will _not_ be synchronized.
> 
> I've only "username" in my list.

只要你用了LDAP登陆方式而且不指定用户组的DN那么每次登陆的时候都会自动同步一次角色到数据库里，除非你在同步选项中把角色同步功能排除出去，文档里就有一模一样的例子。
![image](https://user-images.githubusercontent.com/20925565/100818736-b52f4200-3485-11eb-9938-df6d83d9d66c.png)
求求你们读一下文档吧！！！！

ChristianSwift:
> How should this solve the problem?
[...]

PR #4674, opened 2020-12-02
Dont force role of user if no ldap groups defined

eSkiSo:
We should not force role of user on LDAP logins if there are not Manager/Admin groups defined, return null to get the one from database as before.


eSkiSo:
Fixes #4640

PR #4675, opened 2020-12-05
Added an option to send a copy of all generated e-mails to a bcc address.

hakakou:


app/Core/Mail/Transport/Mail.php
```diff
@@ -36,6 +36,10 @@ public function sendEmail($recipientEmail, $recipientName, $subject, $html, $aut
                 ->setFrom($this->helper->mail->getMailSenderAddress(), $authorName)
                 ->setTo(array($recipientEmail => $recipientName));
 
+            if (defined('MAIL_BCC')) {
```
fguillot, line 39: You should test if the value is not an empty string.

config.default.php
```diff
@@ -44,6 +44,9 @@
 // E-mail address used for the "From" header (notifications)
 define('MAIL_FROM', 'replace-me@kanboard.local');
 
+// Bcc email for a copy of all notifications. Comment out if not needed.
+// define('MAIL_BCC', 'replace-me@kanboard.local');
```
fguillot, line 48: The default value should probably be an empty string and uncommented.

fguillot (commented):
The new constant should also be defined here like others: https://github.com/kanboard/kanboard/blob/master/app/constants.php

hakakou:
Thank you for your directions. I think I've implemented correctly.

hakakou:
Sorry closed by mistake

PR #4676, opened 2020-12-08
grammatical error

G2G2G2G:
This is also kind of appropriate:
I want to receive notifications only for those projects 
->
I want to receive notifications only for these projects

those is to that as these is to this


fguillot (commented):
The translation keys in the locale folder needs to be updated as well.

PR #4677, opened 2020-12-08
Advanced bulk task creation (using text) + subtasks.

maxigit:
Here is what I've do to allow faster way of creating task in bulk.
I suppose the code doesn't meet the required  standard, and it would probably better be in a separate plugin. However
I'm not sure if there are any hooks available to do si. Anyway, here it is, if someone wants to have a look.
I find it really useful.

Allows tags  description and color to be set for each individual task
when using bulk task creation.

The current bulk task creation, create on task per line, and the use the
full line as the task tile.

This commit allows some attributes to be specified in the task line with
the following syntax

    tile [the description] !color :tag1:tag2

It also allows to create subtasks. Each line starting with a `-` will be
added as a subtask of the previous task. A subtask can also contains a
status indicator (similar to orgmode).

Example

    Task1 !blue
    Task2 [Description of task 2]
    Task 3 :tag1:tag2
    - subtask 1
    - [X] already done.
    - [-] in progress.

Create 3  tasks

   Task1 with color set to blue
   Task 2 with the description set to "Description of task 2"
   Task 3 set with the tags, tag1 and tag2.
   Task contains also 3 subtasks, with a different progress status.


Taeluf:
I think the `model:task:creation:prepare` hook might work (see [github.com/kanboard/kanboard/blob/...](https://github.com/kanboard/kanboard/blob/186d34cb4ecf8b8b88e2c5586eff9b387f38eaf2/app/Model/TaskCreationModel.php#L91) if you wanted to go the plugin route. You'd have to check if the requested controller was the TaskBulkController (or allow your syntax with all new task titles)

Another option might be to pr to add hooks to the TaskBulkController, then a plugin could benefit from those hooks.

I'm also wondering if there's a way to hook into the any ole controller call, but I couldn't find one.

Anyway, I think its a great idea & looks nicely implemented!

fguillot:
The plugin route might be better for that kind of changes.

Having all this code in the controller method is not ideal. It should be separated into one or multiple functions to be easily unit tested.

maxigit:
@ReedyBear I think `model:task:creation:prepare` hook might work indeed for the color and task parsing, but I'm not sure it allows the subtasks parsing/creation. I agree this could be two separate plugins, even thought I'm not sure why one would use one but not the other. Anyway to go the plugin route, I think another hook would be needed.

PR #4694, opened 2020-12-29
Update AuthSubscriber.php

ThreeCO:
For Docker with ReverseProxy
- Added IP-Address to error log since it did only show docker-gateway in logs. 
- With this modification fail2ban can be used.

### DIFF
--- a/app/Subscriber/AuthSubscriber.php
+++ b/app/Subscriber/AuthSubscriber.php
@@ -95,10 +95,11 @@ class AuthSubscriber extends BaseSubscriber implements EventSubscriberInterface
     {
         $this->logger->debug('Subscriber executed: '.__METHOD__);
         $username = $event->getUsername();
-
+        $ipAddress = $this->request->getIpAddress();
+        
         if (! empty($username)) {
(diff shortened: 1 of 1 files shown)


app/Subscriber/AuthSubscriber.php
```diff
@@ -95,10 +95,11 @@ public function onLoginFailure(AuthFailureEvent $event)
     {
         $this->logger->debug('Subscriber executed: '.__METHOD__);
         $username = $event->getUsername();
-
+        $ipAddress = $this->request->getIpAddress();
+        
         if (! empty($username)) {
             // log login failure in web server log to allow fail2ban usage
-            error_log('Kanboard: user '.$username.' authentication failure');
+            error_log('Kanboard: user '.$username.' authentication failure with IPAddress: ' $ipAddress);
```
fguillot, line 102: This is not valid in PHP. You must add a dot `.` to concat strings.

And the wording could be changed, for example `IP address:` instead of `IPAddress:`.

fguillot (changes requested):

PR #4719, opened 2021-02-01
LDAP protocol/host/port configuration by URL; make BASE_DN optional

operateur404:
Querying an Active Directory Global Catalog across an entire forest requires : 
- using `ldap_connect($uri)` instead of `ldap_connect($host, $port)`
- using an empty base DN

PHP `ldap_connect($host, $port)` function signature is deprecated : https://www.php.net/manual/en/function.ldap-connect.php

### DIFF
--- a/app/Core/Ldap/Client.php
+++ b/app/Core/Ldap/Client.php
@@ -70,8 +70,8 @@ class Client
      *
      * @access public
      *
-     * @param  string $server LDAP server hostname or IP
-     * @param  int    $port   LDAP port
+     * @param  string $server LDAP server URI (ldap[s]://hostname:port) or hostname (deprecated) 
+     * @param  int    $port   LDAP port (deprecated)
      * @param  bool   $tls    Start TLS
      * @param  bool   $verify Skip SSL certificate verification
      * @return Client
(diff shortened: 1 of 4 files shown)


fguillot:
Please, update the tests accordingly to the changes: https://github.com/kanboard/kanboard/pull/4719/checks?check_run_id=1808119556#step:6:17

operateur404:
Test updated, sorry for the inconvenience.
Thanks and keep up the good work !

app/Core/Ldap/Client.php
```diff
@@ -88,7 +88,12 @@ public function open($server, $port = LDAP_PORT, $tls = LDAP_START_TLS, $verify
             putenv('LDAPTLS_REQCERT=never');
         }
 
-        $this->ldap = @ldap_connect($server, $port);
+        if (is_integer($port)) {
```
fguillot, line 91: The default value of `$port` is `389`, so it will always use the deprecated from of `ldap_connect()` unless `$LDAP_PORT` is set to a non-integer value. 

https://github.com/kanboard/kanboard/blob/master/app/constants.php#L66

What about testing if the `$LDAP_SERVER` value starts with `ldap[s]://` instead?

PR #4722, opened 2021-02-01
Fix path to plugin files when PLUGINS_DIR is something other than DOCUMENT_ROOT/plugins

Taeluf:
Regarding issue https://github.com/kanboard/kanboard/issues/4721

PR #4720, opened 2021-02-01
Use an absolute filepath in AssetHelper css() & js() functions

Taeluf:
Regarding issue https://github.com/kanboard/kanboard/issues/4657

I have an unconventional Kanboard install, where kanboard is in a subdirecory of the DOCUMENT_ROOT. 

This causes error like:
```
 Warning: filemtime(): stat failed for assets/css/vendor.min.css in DOCUMENT_ROOT/kan/app/Helper/AssetHelper.php on line 43
```
where kanboard is in the `kan` folder.

This PR builds an absolute path for the filemtime() call, which fixes the problem.

PR #4723, opened 2021-02-01
Merge filemtimefix andpluginsdir fix

ReedyBear:
This merges my other two PRs:
https://github.com/kanboard/kanboard/pull/4722
and
https://github.com/kanboard/kanboard/pull/4720

Alltogether, these allow a kanboard install to:
- be stored in a subdirectory of the document root
- Use an alternative plugins_dir
- Keep existing plugins working that use the plugins/ assumption

This fails to:
- Provide a standardized way to provide your asset path in a plugin
    - Plugins will still need to use `plugins/MyPlugin/Assets/myasset.css`.

### DIFF
--- a/app/Helper/AssetHelper.php
+++ b/app/Helper/AssetHelper.php
@@ -21,7 +21,9 @@ class AssetHelper extends Base
      */
     public function js($filename, $async = false)
     {
-        return '<script '.($async ? 'async' : '').' defer type="text/javascript" src="'.$this->helper->url->dir().$filename.'?'.filemtime($filename).'"></script>';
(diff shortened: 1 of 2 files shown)


Taeluf:
My apologies (& please advise) if I should have PRd these differently.

fguillot (changes requested):
That seems to be very "hackish". Can you add some unit tests to make sure nothing is going to break?

ReedyBear:
Yeah, it is kind of hackish ... Umm ... I haven't used kanboard in awhile & likely won't work on this any time soon. I'll close the PR. Sorry.

PR #4746, opened 2021-03-03
new hook model:task:duplication:aftersave

BlueTeck:
New hook to allow plugins to do something after task duplication, without overwriting core models.

See https://github.com/BlueTeck/kanboard_plugin_coverimage/pull/33

Do you follow the guidelines?

- [X] There is no regression
- [X] I follow existing [coding style](https://docs.kanboard.org/en/latest/developer_guide/coding_standards.html)

### DIFF
--- a/app/Model/TaskDuplicationModel.php
+++ b/app/Model/TaskDuplicationModel.php
@@ -60,7 +60,7 @@ class TaskDuplicationModel extends Base
             $this->taskLinkModel->create($new_task_id, $task_id, 4);
         }
 
-        $hook_values = [ 'source_task_id' => $task_id, 'target_task_id' => $new_task_id];
+        $hook_values = [ 'source_task_id' => $task_id, 'destination_task_id' => $new_task_id];
         $this->hook->reference('model:task:duplication:aftersave', $hook_values);
 
         return $new_task_id;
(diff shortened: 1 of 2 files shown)


app/Model/TaskProjectDuplicationModel.php
```diff
@@ -33,6 +33,9 @@ public function duplicateToProject($task_id, $project_id, $swimlane_id = null, $
             $this->taskLinkModel->create($new_task_id, $task_id, 4);
         }
 
+        $hook_values = [ 'source_task_id' => $task_id, 'target_task_id' => $new_task_id];
+        $this->hook->reference('model:task:duplication:aftersave', $hook_values);
```
fguillot, line 37: This hook should be named differently that the first one? It's specific to project duplication.

app/Model/TaskDuplicationModel.php
```diff
@@ -60,6 +60,9 @@ public function duplicate($task_id)
             $this->taskLinkModel->create($new_task_id, $task_id, 4);
         }
 
+        $hook_values = [ 'source_task_id' => $task_id, 'target_task_id' => $new_task_id];
```
fguillot, line 63: Other parts of the code use the term "destination" instead of "target".

app/Model/TaskProjectDuplicationModel.php
BlueTeck: changed

app/Model/TaskDuplicationModel.php
BlueTeck: changed

PR #4767, opened 2021-04-04
Added language "Deutsch (du)"

manne65-hd:
Do you follow the guidelines?

- [x] There is no regression

### Introducing: Deutsch (du)
I was missing the "non-formal" version of the german translation ... so here it is.
While making the necessary changes I also found some inconsistent useage of "non-formal" existing german translation, which I also fixed.

In order not to break anything I decided to leave **de_De** untouched, but just change it's name for the dropdown-list to "Deutsch (Sie)", so if anyone has set "Deutsch" as the application-standard or as the language of choice in his profile this will remain intact, with the difference, that it will now list as "Deutsch (Sie)"


app/Model/LanguageModel.php
```diff
@@ -147,6 +149,7 @@ public function getJsLanguageCode()
             'ca_ES' => 'ca',
             'da_DK' => 'da',
             'de_DE' => 'de',
+            'de_DE_du' => 'de-du',
```
fguillot, line 152: The Javascript language should be `de` because jquery-ui doesn't seems to support `de-du`.

manne65-hd: Should be resolved now

Issue #4581, opened 2020-08-24
Automated action when moving task to other column does not trigger if it's moved across swimlakes


MavropaliasG:
### Actual behaviour

I set up a rule for changing color of a task when it's moved to a specific color. When I move it to that column the action triggers, but if I move it to the column but in another swimlake it does not.
### Expected behaviour

The action (colour change) should trigger when moving a task in the specified column regardless if it's in the same or another swimlake

### Steps to reproduce
* Created automated task
* Move task to the column within the same swimlake
* Move task to the column in a different swimlake

### Configuration

Application version: 1.2.15
PHP version: 7.4.8
PHP SAPI: litespeed
HTTP Client: cURL
OS version: Linux 3.10.0-962.3.2.lve1.5.27.el6h.x86_64
Database driver: sqlite
Database version: 3.26.0
Browser: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.125 Safari/537.36

manne65-hd:
Hi Georgios,
I had the same issue and have finally managed to make a pull-request.

https://github.com/kanboard/kanboard/pull/4772

Let's see if it gets approved :-)

MavropaliasG:
Thank you very much @manne65-hd

Issue #4163, opened 2019-03-22
AUTO-ACTION "AssignColor ..." ignored when moving between column and swim lane at once


manne65-hd:
### Actual behaviour
As reported long ago [in this issue ](https://github.com/kanboard/kanboard/issues/1365)... it looks like the problem is still not addressed.
If have a project with 5 columns and 2 swimlanes. In Automatic actions I have setup a couple of "Assign a color when task is moved to a specific column"-actions. These only apply, when a task is moved within the same swimlane.
When moving a task from swimlane A / column A to Swimlane B / column B no action is triggered.

### Expected behaviour
Automatic Actions of the type "Assign a color ..." should be triggered regardless "swimlane-changes" when moving between columns. 

I'm still new to kanboard so I don't have the skills(yet) to try and fix this myself ... but maybe someone can try and point me in the right direction, where to try and fix this ?

### Steps to reproduce
Setup a project with at least 2 swimlane and setup at least on Automatic Action of the type "Assign color when moving ..."
Then move a task within the same swimlane ... color change will apply!
Move another task from swimlane A / column A to swimlane B / column B ... color change is ignored.

### Configuration
- Kanboard version: 1.2.8
- Database type and version: mysql 5.6.29
- PHP version: 7.1.19
- OS: Windows Server 2016 with IIS
- Browser: Chrome 73.0.3683.86

rafacamargo123:
@manne65-hd Since you asked to someone to point you in the right direction, I'll explain to you why this bug exists, then you can try some solutions and maybe request a pull request with the changes. I don't know which one is the best solution below.
This is happening because the automatic action you are using -- Assign a color when task is moved to a specific column -- is handling the events EVENT_MOVE_COLUMN and EVENT_CREATE. This task is executed when this event is fired. You set on which event the task will be executed when you create an automatic action, and this automatic action only supports those events:
https://github.com/kanboard/kanboard/blob/11b6bf6d25668b8bb9d0304c30f803f8194ab1d8/app/Action/TaskAssignColorColumn.php#L32-L38

Automatic action creation:
![image](https://user-images.githubusercontent.com/12586009/57042955-8b27f500-6c3c-11e9-84b1-a24166620713.png)
Now, let's check why when a task is moved between swimlanes and columns this event is not being fired. The event is fired on the Model.
[...]

Issue #1365, opened 2015-10-13
Actions not applied to cards moved between columns in different swimlanes


nmarus:
I searched through the issue log, so my apologies if this is a dupe. 

Moving cards to another column in another swim lane of the same board does not apply the actions. The actions are only applied within the swimlane. 

Board Setup:
- 4 swim lines (S1,S2,S3,S4)
- 4 columns (C1,C2,C3,C4)

Actions (8):
- change color when moving card to a specific column (1 for each col)
- change color when creating card in specific column (1 for each col)

Scenario:
- Card1 in S1C1 -> S1C2 -> S1C3, etc (actions work)
- Card1 in S1C1 -> S2C2 (actions do not work)

sparkles645:
@nmarus 
Tested, with the same results.

Automatic action -> "Assign a color when the task is moved to a specific column"

When moving task within the same swimlane, works fine.
When moving task directly to a different swim lane, **does not work**.

fguillot:
It doesn't work because that trigger the event `task.move.swimlane`. That need to be improved or new action may be created to handle that.

bugwatcher:
This issue has been closed automatically by a bot for inactivity after more than 1 year. If you still need this feature, you should consider contributing to this open-source project.

PR #4772, opened 2021-04-05
Task pos model fire all events

manne65-hd:
Do you follow the guidelines?

- [x] There is no regression

### Fire ALL events, when moving tasks between columns AND swimlanes _(configurable via config.php)_
There have been 3 reports in the past, about the problem that the automatic action _"Assign a color when the task is moved to a specific column"_ will only work, if the task is moved within the same swimlane.

- https://github.com/kanboard/kanboard/issues/1365
- https://github.com/kanboard/kanboard/issues/4163
- https://github.com/kanboard/kanboard/issues/4581

This PullRequest could fix these issues

#### Why is this NOT a breaking change ?
By default everything will work as before, but Administrators now have the ability to configure which EVENTS will be triggered, when tasks are being moved between swimlanes AND columns at the same time.

#### How does it work?
By setting the new parameter TRIGGER_TASK_MOVE_EVENTS in **config.php** you can control the new behavior:
- '' _(default)_
  - ONLY the MoveSwimlane-Event will be triggered _(old behaviour)_
- 'Column' 
  - ONLY the MoveColumn-Event will be triggered
- 'SwimLane+Column'
  - the MoveSwimlane-Event AND the MoveColumn-Event will be triggered
  - BEWARE >> This will also trigger 2 seperate Mail/Web-Notifications for users that subscribe to that project
- 'All'
  - the MoveSwimlane-Event, MoveColumn-Event AND the MovePosition-Event will be triggered
  - BEWARE >> This will also trigger 3 seperate Mail/Web-Notifications for users that subscribe to that project

#### Testing ?
I did quite some (manual) testing on my local machine, but if this requires adding unit-tests ... I might need help, because this is still beyond my knowledge 

Hope you like it


manne65-hd:
Just a humble **bump** ...
Do you have any thoughts on this one?

The reason why I added so "many" options, is the fact that depending on which of the new options an ADMIN chooses this involves triggering more mails. So it's up to the one running a kanboard-instance to decide, what's more important to them.

 Sum up all possible change-event-notifications in a single mail ??
-----------------------------------------------------------
I'd be willing to investigate the possibility to change the mail-notifications in order to "collect" all possible 3 changes and notify them in a single mail. 
But as my understanding of kanboard's code is still at a beginner-level, I have no idea how complicated that would be and also if you would even be willing to accept such (breaking) changes.

manne65-hd:
@fguillot 
Just another humble bump on this one ...
Could you please give me some feedback on this pull-request?

Looking forward to your answer,
Manfred

fguillot:
I'm not really fan of `TRIGGER_TASK_MOVE_EVENTS`. Average users will keep the default value and that might introduce confusion for those using a different value. Triggering multiple notifications might create too much noise.

I would rather use a different approach. Perhaps, introducing a new event for tasks moved to a different swimlane and a different column, assuming the intend is to fix the automatic action mentioned above. I haven't touched this code in years, so I might be wrong.

manne65-hd:
> I'm not really fan of `TRIGGER_TASK_MOVE_EVENTS`. Average users will keep the default value and that might introduce confusion for those using a different value. Triggering multiple notifications might create too much noise.

I understand that concern

> I would rather use a different approach. Perhaps, introducing a new event for tasks moved to a different swimlane and a different column, assuming the intend is to fix the automatic action mentioned above. I haven't touched this code in years, so I might be wrong.

Am i right in assuming, that this would require to create twice the amount of automatic actions?
 - Action to execute when EVENT_MOVE_COLUMN or EVENT_MOVE_SWIMLANE was triggered _(as in current version of Kanmboard)_
   - "Same" Action to execute when new EVENT_MOVE_COLUMN_SWIMLANE was triggered ?

manne65-hd:
> I would rather use a different approach. Perhaps, introducing a new event for tasks moved to a different swimlane and a different column, assuming the intend is to fix the automatic action mentioned above. I haven't touched this code in years, so I might be wrong.

OK ... I gave this a try and introduced a new EVENT_MOVE_COL_AND_SWL.
From what I have tested locally it looks to me like to works ... but I will have to give up on trying to write a unit-test for this, as this is simply lightyears out of the scope of my (current) skill-set ;-/

Am willing to learn though, but it looks like there are many changes required to be able to run successful unit-tests on this new event.

Please advice how to go on from here.

Thank you,
Manfred

fguillot:
Closing stale pull-request

PR #4786, opened 2021-04-18
Add a method to get the latest comment for a given for task

manne65-hd:
Do you follow the guidelines?

- [x] There is no regression


fguillot (commented):
Where is it used?

If it's for a plugin then it might be more appropriate to move this function to your plugin.

manne65-hd:
> If it's for a plugin then it might be more appropriate to move this function to your plugin.

I am in fact using it in a plugin under development.
My humble idea was, that _(since it's so small)_ it could also be useful in the core-code.

But you're right ... I can also just add it to my plugin, if you don't feel there's any future use for this in kanboard's core

manne65-hd:
Will add the new method within my plugin

Issue #4759, opened 2021-03-20
First swimlane title should be above the column names for UI consistency [Screenshot examples included]


groovybluedog:
When having a project with multiple swimlanes, on the first swimlane, the column names are above and then the swimelane name below. However for all other swimlanes, the swimlane name is above and the column name is below.

![image](https://user-images.githubusercontent.com/16142383/111856795-2277bd80-8925-11eb-89ac-3c8c81c104f4.png)

I find this UI choice a bit odd, and this seemingly minor difference is inconsistent and can be confusing to new users. 

I feel that, when there are multiple swimlanes, the first one should replicate the others and have the column name below, which should also hide when the swimlane name is collapsed. I do not feel that there is a need to have a "sticky" column name at the top when it always shows under all other swimlanes anyway.

This is my bad example in Paint of a more consistent UI when swimlanes are involved.

![image](https://user-images.githubusercontent.com/16142383/111856949-28ba6980-8926-11eb-8254-51bbf38d28f0.png)

- [ AFFIRMATIVE ] I have read this document: https://docs.kanboard.org/en/latest/developer_guide/faq.html#why-are-you-not-developing-my-feature-request

manne65-hd:
@groovybluedog 
I am also facing the issue, that my new users find this confusing ... however after my investigations, the column-header must be rendered first in order for the **collapse-column-feature** to work correctly.
So I decided to go with the seond-best attempt and make introduce a consistent behaviour _(column-header first / swimlane-header second)_ for all swimlanes.

Let's see if my Pull-request gets accepted ... otherwise I consider releasing this as a plugin.

PR #4791, opened 2021-04-20
Consistent Column+Swimlane-Headers

manne65-hd:
Do you follow the guidelines?

- [x] There is no regression

(Sort of) fixes: https://github.com/kanboard/kanboard/issues/4759

As reported in the issue above, I have also experienced that our users find the inconsistent order of the column and swimlane-headers slightly confusing.
From what I know for now, the columns need to be rendered first to "generate" the general setup of columns in order for the collapsing of columns to work correctly. So my attempt just makes the column-headers always appear above the swimlane-header.
The swimlane-header is still being suppressed, if only one swimlane exists.

Hope you like it.

### DIFF
--- a/app/Template/board/table_container.php
+++ b/app/Template/board/table_container.php
@@ -21,8 +21,14 @@
         <?php foreach ($swimlanes as $index => $swimlane): ?>
             <?php if (! ($swimlane['nb_tasks'] === 0 && isset($not_editable))): ?>
 
-                <!-- Note: Do not show swimlane row on the top otherwise we can't collapse columns -->
-                <?php if ($index > 0 && $swimlane['nb_swimlanes'] > 1): ?>
+                <!-- Note: Columns must render first otherwise we can't collapse them -->
+                <?= $this->render('board/table_column', array(
+                    'swimlane' => $swimlane,
+                    'not_editable' => isset($not_editable),
+                )) ?>
+
+                <!-- Hide first swimlane-header if project has only 1 swimlane -->
+                <?php if ($index === 0 && $swimlane['nb_swimlanes'] > 1): ?>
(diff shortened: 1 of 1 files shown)


fguillot:
Having the swimlane headers above the column headers is better but HTML tables seems limited. CSS grids might offer more flexibility these days.

groovybluedog:
@manne65-hd Thank you very much for your reply on my issue and making this change, which I can't do myself as I'm not a programmer. This has upped the swimlane usability of Kanboard tenfold. That being said, having the column titles above the swimlane name still looks odd and I feel like it should be the other way around, which is hopefully something the developers can look at changing officially in the future to make the swimlane UI top class. Thanks again, tried a ZIP download and much happier with this change.

manne65-hd:
@fguillot 
> Having the swimlane headers above the column headers is better but HTML tables seems limited. CSS grids might offer more flexibility these days.

I have made [a new attempt to solve this](https://github.com/kanboard/kanboard/pull/4793) in a way, where swimlane-headers can always appear on top of the column-headers ...

 I was so focused on seeing the success on swimlanes appearing on top of columns and the (working) ability of collapsing/uncollapsing swimlanes .... that I first didn't realize, that the collapsed columns don't "shrink" 
 
 I might need some assistance to find the additonal CSS and/or JS-changes to be made, in order for it to **really** work.

Maybe you can help?

manne65-hd:
> That being said, having the column titles above the swimlane name still looks odd and I feel like it should be the other way around, which is hopefully something the developers can look at changing officially in the future ...

@groovybluedog ... I have good news:
[The future is here](https://github.com/kanboard/kanboard/pull/4793)  ;-)

PR #4812, opened 2021-05-12
Fix word and add style cancel button

lfandrade:
Fix word and add style cancel button


fguillot (commented):
You need to include your modifications in the non minified CSS file.

See https://docs.kanboard.org/en/latest/developer_guide/assets.html

lfandrade:
> You need to include your modifications in the non minified CSS file.
> 
> See https://docs.kanboard.org/en/latest/developer_guide/assets.html

Hi, it's done ! Thank you for your reply.

lfandrade:
Hi, ready for merge

fguillot:
Your stylesheet doesn't really look consistent with the rest of the application. I consider that to be a breaking change.

image

It's probably better to use the custom stylesheet in that case.

Your pull-request contains 2 unrelated changes: a translation update and stylesheet change.

PR #4832, opened 2021-06-07
WIP - Add bidi support

ahangarha:
The objective of this WIP PR is to add support for:
- [x] set page direction automatically based on the selected lanugage
- [ ] add bidi support, so that regardless of the selected language, texts used in cards appear in relevant direction and alignment

Do you follow the guidelines? (not checked while in WIP stage)

- [x] There is no regression

### DIFF
--- a/app/Template/layout.php
+++ b/app/Template/layout.php
@@ -50,6 +50,7 @@
           data-timezone="<?= $this->app->getTimezone() ?>"
           data-js-date-format="<?= $this->app->getJsDateFormat() ?>"
           data-js-time-format="<?= $this->app->getJsTimeFormat() ?>"
+          <?php if ($this->app->isRTL($this->app->jsLang())) {echo('dir="RTL"');} ?>
     >
 
     <?php if (isset($no_layout) && $no_layout): ?>
--- a/app/Helper/AppHelper.php
+++ b/app/Helper/AppHelper.php
@@ -195,4 +195,31 @@ class AppHelper extends Base
 
         return '';
     }
+
+    /**
+     * Chech if the given language is a Right-to-Left Language
+     *
+     * @access public
+     * @param  string $lang
+     * @return boolean
+     */
+    public function isRTL(String $lang = '')
+    {
+        $rtl_languages_list = [
+            'ar',
+            'arc',
+            'dv',
+            'fa',
+            'ha',
+            'he',
+            'khw',
+            'ks',
+            'ku',
+            'ps',
+            'ur',
+            'yi'
+        ];
+
+        return in_array($lang, $rtl_languages_list);
+    }
 }


app/Template/layout.php
```diff
@@ -50,6 +50,7 @@
           data-timezone="<?= $this->app->getTimezone() ?>"
           data-js-date-format="<?= $this->app->getJsDateFormat() ?>"
           data-js-time-format="<?= $this->app->getJsTimeFormat() ?>"
+          <?php if ($this->app->isRTL($this->app->jsLang())) {echo('dir="RTL"');} ?>
```
fguillot, line 53: HTML has `dir="auto"` attribute that might simplify this.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir

ahangarha: In this case, it doesn't. `dir="auto"` works with first strong character on an element. This would work for individual elements (on some of which I should add this) but not necessarily on body tag. Here my intention was to force RTL for layout if the language is one of RTL languages like Farsi (Persian) or Arabic.

ahangarha:
I cannot run unittest for this project even if it is a fresh clone. This is what I get:
```bash
✗ phpunit -c tests/units.sqlite.xml
PHP Fatal error:  Declaration of Base::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in /PATH/kanboard/tests/units/Base.php on line 23
```

app/Template/layout.php
ahangarha: So far, everything looks neat. There is only one minor issue for markdown editor which wouldn't occur in regular situations. t should be resolved at the level of markdown parser (by adding `dir="auto"` to each direct child element). I am not sure if we can do it without modification in parser or using js for DOM manipulation.

The other issue is with one class which is hardcoded for floating to right side. Not sure what to do with it. Should I change the class name and then its styles?

When do you think this PR would be ready for merge (if it would be at all)?

fguillot:
Your changes are breaking the page layout when selecting English language:

image

When selecting the Persian language, all menus are on the opposite side, is it what you expect? I thought only the text was right-to-left?

image

ahangarha:
May I know the exact issue with layout if English language be selected.

That time I checked every scenario I could imagine including RTL text in LTR layout and the opposite.

fguillot:
> May I know the exact issue with layout if English language be selected.

Compare the page layout without your changes:

image

And with your changes:

image

ahangarha:
I see the setting icon occupy whole row. I will look into it.

Thank you for reporting.

fguillot:
Are you planning to continue to work on this PR? or I should consider this to be abandoned?

ahangarha:
> Are you planning to continue to work on this PR? or I should consider this to be abandoned?

Hi @fguillot. Thank you for reminding me about this project. I am sorry for the fact I entirely forgot it. I try to dedicate some time to finalize this PR.

ahangarha:
I started working on this issue and noticed that the issue only appears on chrome based browsers.

I am working on it.

ahangarha:
Please check project now.

Still I can make improvement but it needs some more deep surgeries.

Markdown parser needs improvement to add better support for bidirectional text. In some cases, some major html/css rewrite is needed for which I prefer to work on a overall UI enhancement if another developer can spend time with me. I want another person who knows the system and the deeper impact of any change to watch the changes. This should be another PR.

I think this PR is ready to be merged. This way we can get feedback. There shouldn't be any breaking but we may get feedback for improvement of bidi support.

ahangarha:
Any update?

ahangarha:
Do you think we should do this step by step, component by component?
This way, reviewing and debugging would become much more straightforward.

ahangarha:
I totally understand.

And I think we have a clear path to take. Now the objective is to add bidi support partially and gradually.

PR #4863, opened 2021-07-29
Change the dropdown menu on hover

YUII09:
For issue #4806 it needs a hoverable dropdown menu so I have makes a change in the app.min.css


fguillot (changes requested):
You have modified the minified file instead of the source file.

See: https://docs.kanboard.org/en/latest/developer_guide/assets.html

PR #4864, opened 2021-07-30
change links styling and add 2 currencies

yongyouth7:
Do you follow the guidelines?

- [/] I have tested my changes
- [/] There is no breaking change
- [/] There is no regression
- [/] I follow existing [coding style](https://docs.kanboard.org/en/latest/developer_guide/coding_standards.html)


app/Helper/ModalHelper.php
```diff
@@ -30,7 +30,7 @@ public function confirmButtons($controller, $action, array $params = array(), $s
             'url'         => $this->helper->url->href($controller, $action, $params, true),
             'submitLabel' => $submitLabel ?: t('Yes'),
             'orLabel'     => t('or'),
-            'cancelLabel' => t('cancel'),
+            'cancelLabel' => t('Cancel'),
```
fguillot, line 33: Why do you change this? This is going to break all translations.

fguillot (changes requested):
Your pull-request contains unrelated changes.

Also, you have modified the minified file instead of the source file.

https://docs.kanboard.org/en/latest/developer_guide/assets.html

PR #4865, opened 2021-07-30
Change styling for navigation bar and sidebar

CYTeo:
Do you follow the guidelines?

- [/ ] I have tested my changes
- [/ ] There is no breaking change
- [/ ] There is no regression
- [/ ] I follow existing [coding style](https://docs.kanboard.org/en/latest/developer_guide/coding_standards.html)


fguillot (changes requested):
You have modified the minified file instead of the source file.

See: https://docs.kanboard.org/en/latest/developer_guide/assets.html

Issue #4409, opened 2020-02-13
Converting subtask to tasks creates wrong internal link direction


double-p:
### Actual behaviour
New task is a "parent of"

### Expected behaviour
New task is a "child of".

I think attributing a former subtask as "parent" is upside down from my POV.

### Steps to reproduce
Convert a subtask to a task, check internal links.

### Reference
#2908 added the idea, implementation likely in
Model/SubtaskTaskConversionModel.php 

### Configuration

Application version: 1.2.13
PHP version: 7.3.14
PHP SAPI: fpm-fcgi
HTTP Client: cURL
OS version: Linux 3.10.0-1062.4.1.el7.x86_64
Database driver: mysql
Database version: 5.7.24

fguillot:
Tested with version 1.2.19. The relationship seems correct to me.

Main task:

![image](https://user-images.githubusercontent.com/323546/120911928-eb6d9a00-c63f-11eb-89d3-0e864ce5ba4c.png)

Subtask converted to task:

![image](https://user-images.githubusercontent.com/323546/120911933-f1fc1180-c63f-11eb-8220-7325394e67ca.png)

davouch:
I am also having trouble with the same behavior as @double-p and cannot replicate the correct behavior @fguillot presents bellow. (v1.2.20 here)

A subchild converted from a main task gains an internal link status of a parent, and the main task gets to be a child of the converted subtask.
![image](https://user-images.githubusercontent.com/1025752/134486446-2e7d66d7-bf41-4c1e-91df-c2e0d5f3cb05.png)

![image](https://user-images.githubusercontent.com/1025752/134486761-1f42eadb-3875-4dc0-b575-e730aa17c3ca.png)

davouch:
And I believe I could have found the source of the problem in the link labels:
![image](https://user-images.githubusercontent.com/1025752/134487303-d982dde2-5ed8-4447-813f-eb80029eaacd.png)

Not sure if this helps, but I am using:
![image](https://user-images.githubusercontent.com/1025752/134487394-bd32d89e-7807-4a5b-923d-2e691bf70148.png)

PR #4888, opened 2021-09-23
Update SubtaskTaskConversionModel.php

davouch:
I (very carefully) suggest, there might be a wrong `link_id` of `6` passed onto the `create` function of the converted subtask, as the link of ID `#6` according to the `links table` in the db belongs to a `is a parent` link label while the correct one for a subtask should be `is a child of` (which it seems to be the ID `7`).

Rlated issue: #4409

Do you follow the guidelines?

- [x] There is no regression

### DIFF
--- a/app/Model/SubtaskTaskConversionModel.php
+++ b/app/Model/SubtaskTaskConversionModel.php
@@ -39,7 +39,7 @@ class SubtaskTaskConversionModel extends Base
         ));
 
         if ($task_id !== false) {
-            $this->taskLinkModel->create($task_id, $subtask['task_id'], 6);
+            $this->taskLinkModel->create($task_id, $subtask['task_id'], 7);
             $this->tagDuplicationModel->duplicateTaskTags($parent_task['id'], $task_id);
             $this->subtaskModel->remove($subtask_id);
         }


davouch:
Additionally, I've checked the initial db configurations in `/App/Schema` and it seems, there is a inconsistency in the order of the `is a child of` and `is a parent of` between databases. While [Mysql.php](https://github.com/kanboard/kanboard/blob/e5d22682cbc13bc6345183d4c3e68f3be3c074aa/app/Schema/Mysql.php#L1067-L1118) defines the link label `is a parent of` as ID 6 and `is a child of` as ID 7, both the [Postgre.php](https://github.com/kanboard/kanboard/blob/e5d22682cbc13bc6345183d4c3e68f3be3c074aa/app/Schema/Postgres.php#L944-L978) and [Sqlite.php](https://github.com/kanboard/kanboard/blob/e5d22682cbc13bc6345183d4c3e68f3be3c074aa/app/Schema/Sqlite.php#L904-L938) work with an inverted order of the two, and define `is a parent of` as ID 7 while `is a child of` is ID 6.

I am sorry for not seeing this through before the pull request, as I am new to the project and am using Mysql, thus did not check the latter two DBs. This would also explain, why the initial issue #4409 was closed by @fguillot as he supposedly was not using MySQL, but one of the other two DB options, thus was unable to replicate the bug.

app/Model/SubtaskTaskConversionModel.php
```diff
@@ -39,7 +39,7 @@ public function convertToTask($project_id, $subtask_id)
         ));
 
         if ($task_id !== false) {
-            $this->taskLinkModel->create($task_id, $subtask['task_id'], 6);
+            $this->taskLinkModel->create($task_id, $subtask['task_id'], 7);
```
fguillot, line 42: The person that did this change hardcoded the link ID. See this commit https://github.com/kanboard/kanboard/commit/d4749a44025b1518cd8f83941f960e2f699b69b7

Your change will work for MySQL, but it's going to break for Sqlite and Postgres users.

The right way of fixing this is to fetch the link ID by using the name. There are few methods available in LinkModel class: https://github.com/kanboard/kanboard/blob/master/app/Model/LinkModel.php

fguillot (changes requested):

app/Model/SubtaskTaskConversionModel.php
davouch: > The right way of fixing this is to fetch the link ID by using the name. There are few methods available in LinkModel class: https://github.com/kanboard/kanboard/blob/master/app/Model/LinkModel.php

Thanks, sounds legit! A bit above my skill grade, I am afraid.

PR #4910, opened 2021-11-23
don't look for project_id for files

Chaosmeister:
it is only used for late accessibility checking (it was already checked in middleware).

With this, you can create stable file links (as long as the file exists)

I need this change for my [inline image plugin](https://github.com/Chaosmeister/PITM)

- [x] There is no regression


xekon:
I tested this and the PITM plugin, works great and would close this: https://github.com/BlueTeck/kanboard_plugin_bigboard/issues/26

Issue #4917, opened 2021-12-03
Footer Icons (task-board-icons-row div) option to display Right Aligned in header (task-board-header div)


xekon:
Currently the div below "task-board-header" is "dropdown" which has the task number and a dropdown menu. This is naturally left aligned.

I have a TON of wasted space on this line, I would absolutely LOVE it if the Footer Icons could be right aligned in the header of cards.

This would make more efficient use of my screenspace!

There is an addon called BoardCustomizer, it has the ability to hide ALL icons, the appearance is very nice because in most cases the card becomes more compact:

![2021-12-02_18-11](https://user-images.githubusercontent.com/106644/144533505-a4cdc306-d986-42dc-88e2-a4f3da3334b2.png)

with this option I could change this:

![2021-12-02_18-09](https://user-images.githubusercontent.com/106644/144533536-4828548e-4358-4946-be72-d06b7a502c3d.png)

to this:

![2021-12-02_18-20](https://user-images.githubusercontent.com/106644/144533714-62383a19-3f1c-4277-a651-8524fba863c3.png)

- [x] I understand that Kanboard is in maintenance mode. It doesn't mean it's abandoned, but there is no significant feature development.
- [x] I have read this document: https://docs.kanboard.org/en/latest/developer_guide/faq.html#why-are-you-not-developing-my-feature-request

PR #4929, opened 2021-12-19
Add Divs and Hooks to accomodate additional layout customization.

Jieiku:
Almost Closes: https://github.com/kanboard/kanboard/issues/4917  Just need to figure out out the best way to handle the `render('board/task_footer'` block (more on that below)

Regarding this block:
```css
.task-board-icons-bottom{
    margin-top:7px
}
```

At some point somebody added this:
```css
.task-board-icons{margin-top:7px}
```
to the minified css without also including it in the source css files.
That is why this margin value is being added in this pull request, to keep the appearance as it currently is.

One of the goals of this pull request is to allow displaying the footer icons in the void of space available in the header.
This will give a more compact layout if somebody chooses to opt for such a layout.

If you move this block from the div task-board-icons-bottom to the div task-board-icons-top then you get a more compact layout.

```php
                <?= $this->render('board/task_footer', array(
                    'task' => $task,
                    'not_editable' => $not_editable,
                    'project' => $project,
                )) ?>
```

I did not include anything in this pull request to facilitate moving this block of code between the two locations.
I was not sure how best handle this, so I decided to leave it for a future pull request and simply get the framework in place with this pull request.
**(help or suggestions on how to implement moving this block of code between the two divs appreciated)**

I was thinking just include the `render('board/task_footer'` block in both the top and bottom icons div and surround them with an if statement that checks if the user is configured to display them in either the top or the bottom.

Here is the stock appearance, it remains unchanged with this pull request:

If you move the footer icons render block into the div task-board-icons-abs-topright, then you get this appearance:

The nested class task-board-icons-abs-topright is so that the icons-top div does not block text and waste space, it is used to solve the problem illustrated in this picture:

This is with the The nested class task-board-icons-abs-topright:

The endgame goal is something similar to this, once this pull request is merged I will work on pull requests to extend the plugins involved (group_assign, bigboard):

Do you follow the guidelines?

- [x] There is no regression


Jieiku:
Closed, because I was unaware that I could override the layout by overriding the template. These changes can be made as part of a plugin, which I am just about finished with.

PR #4953, opened 2022-02-13
Removed font-family specification for input and textarea

vedranmiletic:
This avoids the requirement to use `!important` in custom CSS to override fonts for input and textarea.

**Note:** this might not be the preferred way to fix the issue as it changes the default fonts (without override) to whatever the user agent has as a default.

Do you follow the guidelines?

- [ ] There is no regression


fguillot:
Unfortunately it's not something that I can merge because it breaks the actual design. Chrome shows a monospace font for textarea with your changes.

vedranmiletic:
Exactly, that's why I did not check "There is no breaking change" and "There is no regression". I am open to suggestions for a better approach.

fguillot:
Can you please provide more context about the problem you are trying to solve?

vedranmiletic:
I'm trying to allow font overrides without `!important`.

vedranmiletic:
Would a pull request that changes the two font-families to `system-ui, sans-serif` be acceptable? CSS font-family: system-ui is [well supported](https://caniuse.com/font-family-system-ui) by now.

aljawaid:
> I'm trying to allow font overrides without `!important`.

where are you trying to target?

vedranmiletic:
> > I'm trying to allow font overrides without `!important`.
> 
> where are you trying to target? In a plugin, you dont need the `!important` it seems
> 
> image

There's `Custom CSS` in `Application settings` dialog.

aljawaid:
> This is not how you change fonts in Kanboard; there's `Custom CSS` in `Application settings` dialog.
> 
> But that was the initial plan. Adding `system-ui`, perhaps with some other modern default fonts (perhaps copying the list from Bootstrap), would work equally well if not better.

I was just showing you the example, if you set the font through a plugin it will override.  I didn't test the custom CSS section.  It was just a mere example for you.

vedranmiletic:
> @vedranmiletic The change below should work. If you don't mind updating your pull-request.

Just tested it on my instance and it seems to work. I have updated the pull request accordingly. Thanks!

PR #4954, opened 2022-02-14
Restore PHP 7.3 support

lazka:
The last update bumped two dependencies leading to newer PHP requirements.
psr/container to 2.0 which requires PHP 7.4 and phpunit to 9.5 requiring
PHP 7.3.

This sets the targeted platform to 7.3 again, so a composer update should
downgrade psr/container and at least restore 7.3 support.

TODO: This is missing a lock file and vendor update.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,10 +11,13 @@
     "config": {
         "preferred-install": "dist",
         "optimize-autoloader": true,
-        "discard-changes": true
+        "discard-changes": true,
+        "platform": {
+            "php": "7.3"
+        }
     },
     "require" : {
-        "php" : ">=7.2.0",
+        "php" : ">=7.3",
         "ext-simplexml" : "*",
         "ext-dom" : "*",
         "ext-xml" : "*",


lazka:
Alternatives are to require PHP 7.4, or also downgrade PHPUnit to 8.x and restore PHP 7.2 support.

vedranmiletic:
Documentation claims that PHP 7.2 is still supported. Perhaps it should be updated as well.

lazka:
I'd be happy to update the docs as well. I just need a PHP version we want to target :)

mikeperalta1:
I respectfully request the minimum requirement be 7.2. That's the default on my RockyLinux install, so I believe this could affect many RHEL/compatible users. I'm able to upgrade to 7.4 but that breaks my imagemagick for some reason, so I prefer to stay with 7.2 for now, if possible. I imagine others might have the same issue.

```
PHP 7.2.24 (cli) (built: Oct 22 2019 08:28:36) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.24, Copyright (c) 1999-2018, by Zend Technologies
```

lazka:
> I respectfully vote the minimum requirement be 7.2. That's the default on my RockyLinux install,

So you expect Kanboard to support PHP 7.2 for 7 more years?

rautamiekka:
> I respectfully vote the minimum requirement be 7.2. That's the default on my RockyLinux install

Surely it's possible to install PHP >7.2 to RockyLinux ? Something that should be done anyway.

lazka:
Thanks!

PR #5025, opened 2022-06-08
Column sorting improvements

Chaosmeister:
extract sorting to separate icon.
add sorting by id

Do you follow the guidelines?

- [x] There is no regression


Chaosmeister:
I'm not quite sure, why the postgres test would fail here 🤔

Chaosmeister:
I added a bit of css that pulls the scores to the right of each column header:

image

PR #5044, opened 2022-08-16
get version from file in app/ directory

duiffie:
instead of getting a version number from the root directory, get it from the app/ directory

Do you follow the guidelines?

- [v] I have tested my changes
- [v] There is no breaking change
- [v] There is no regression
- [v] I follow existing [coding style](https://docs.kanboard.org/en/latest/developer_guide/coding_standards.html)

### DIFF
--- a/app/functions.php
+++ b/app/functions.php
@@ -198,8 +198,8 @@ function build_app_version($ref, $commit_hash)
 
     if ($commit_hash !== '$Format:%H$') {
         return 'main.'.$commit_hash;
-    } else if (file_exists('/version.txt')) {
-        return rtrim(file_get_contents('/version.txt'));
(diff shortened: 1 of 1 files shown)


duiffie:
> What problem does it solve?

When not cloning the git repository but (for instance) creating a custom kanboard package the version is not shown in the settings page. When viewing the code I saw that it was already possible to use a version.txt. But it seemed a bit odd to me to place that (specific kanboard) file in / on the server. The kanboard app/ directory seems more suitable to me.

duiffie:
> The Dockerfile should be updated accordingly: https://github.com/kanboard/kanboard/blob/main/Dockerfile#L22

Done

PR #5065, opened 2022-09-15
More precise zh_CN translations

greyaz:
Do you follow the guidelines?

- [ ] There is no regression
- [ ] I follow existing [coding style](https://docs.kanboard.org/en/latest/developer_guide/coding_standards.html)


greyaz:
The failure seems nothing to do with the updates of the translation file, please have a look. @fguillot

PR #5078, opened 2022-10-10
tests: port assertEquals() with delta to assertEqualsWithDelta()

lazka:
assertEquals() in phpunit 9.5 no longer takes a delta parameter and has assertEqualsWithDelta() as a replacement. This means float get compared without a delta atm, and a recent phpunit release (9.5.25) has made float comparisons stricter resulting in test suite errors such as:

1) SubtaskTimeTrackingModelTest::testCalculateSubtaskTime Total spent
Failed asserting that 3.3000000000000003 matches expected 3.3. tests/units/Model/SubtaskTimeTrackingModelTest.php:186

This replaces all assertEquals() calls that pass a delta value with assertEqualsWithDelta().

Do you follow the guidelines?

- [x] There is no regression

### DIFF
--- a/tests/units/Action/TaskUpdateStartDateTest.php
+++ b/tests/units/Action/TaskUpdateStartDateTest.php
@@ -36,7 +36,7 @@ class TaskUpdateStartDateTest extends Base
 
         $task = $taskFinderModel->getById(1);
         $this->assertNotEmpty($task);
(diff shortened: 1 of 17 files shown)


lazka:
For extra context here is the recent phpunit change resulting test errors: https://github.com/sebastianbergmann/phpunit/pull/4972

lazka:
Thanks!

PR #5037, opened 2022-07-13
Add support for MS SQL Server as a database backend for Kanboard

jnahmias:
Do you follow the guidelines?

- [X] I have tested my changes
- [X] There is no breaking change
- [X] There is no regression
- [X] I follow existing [coding style](https://docs.kanboard.org/en/latest/developer_guide/coding_standards.html)

PR #5082, opened 2022-10-24
Add support for Microsoft SQL Server database

jnahmias:
Do you follow the guidelines?

- [X] There is no regression
- [X] I follow existing [coding style](https://docs.kanboard.org/en/latest/developer_guide/coding_standards.html)

This is the third and final set of commits towards enabling kanboard to use MSSQL for its database backend, as discussed in PR #5037. These changes build on the previous two batches.
Here we fix a couple of issues with the tests as written; add the MSSQL database schema, provider, and app configuration; add support in the unit test suite for two of the MSSQL pdo drivers; skip a failing test on ODBC; and finally, run the unit tests on MSSQL/ODBC as a GitHub Action.


jnahmias:
@fguillot - If you wouldn't mind, I think it would be good if you would make a new version release once you've merged these patches. Thanks again!

PR #5088, opened 2022-11-02
Issue with #4830

creecros:
The added `.$size`  in `'hash' => md5($user['avatar_path'].$size` appears to be incorrectly showing avatar images.
image

Not sure if it was intentional or not, but it appears to be a mistake, removal fixes the issue.

Do you follow the guidelines?

- [x] There is no regression


creecros:
wiped the container and tested again, now i have no issue. i thought i did that last time but maybe i didn't. must be getting old.

creecros:
And its not customizer plugin causing it. 

Although, this does make my slider preview for changing the avatar size more challenging, since it was all done in js. 

yikesfruitstripesgum

all good, i think i know what i can do to fix.

PR #4652, opened 2020-10-26
Display number of tasks according to filter

rhorber:
This resolves #4582.

To the total number of tasks, and if set task limit, additionally the displayed number of tasks is shown.

Feel free to reject the PR if you do not want this feature (due to BC or regressions).

PR #5108, opened 2022-12-14
Update the wrong number of tasks in the column header

greyaz:
The current value of `$column['column_nb_open_tasks']` represents the number of open tasks in the current column of a swimlane.
But according to the description: "Total number of tasks in this column across all swimlanes", the value of `$column['column_nb_tasks']` should be more proper.

Do you follow the guidelines?

- [x] There is no regression


greyaz:
Hi @fguillot ,
I believe the behaviour of the PR #4652 is not correct. Please check the screenshots I made in my test environment based on current behaviour.
But my previous commitment also needs to be corrected. I missed the scenario of adding extra filters and made it up in the new one.

Pasted Graphic 3
Pasted Graphic 4

PR #5115, opened 2023-01-04
Fix DateTime::getLastErrors return validation

AH-DL:
Do you follow the guidelines?

- [X] There is no breaking change :

Simple modification following DateTime extension docs, the return value of the method is always false when no error or warnings are stored, then the variable "$errors" should not be evaluated as an array inside the conditional.

[PHP Documentation] (https://www.php.net/manual/en/datetime.getlasterrors.php)

- [X] There is no regression
- [-] I have updated the unit tests and integration tests accordingly

### DIFF
--- a/app/Locale/ru_RU/translations.php
+++ b/app/Locale/ru_RU/translations.php
@@ -495,7 +495,7 @@ return array(
     'is duplicated by' => 'дублирует',
     'is a child of' => 'является продолжением',
     'is a parent of' => 'является началом для',
-    'targets milestone' => 'часть вехи',
+    'targets milestone' => 'нацелена на веху',
     'is a milestone of' => 'является вехой для',
     'fixes' => 'исправлено',
     'is fixed by' => 'исправляет',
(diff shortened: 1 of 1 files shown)


app/Core/DateParser.php
```diff
@@ -204,13 +204,13 @@ private function getValidDate($value, $format)
...
             $errors = DateTime::getLastErrors();
-            if ($errors['error_count'] === 0 && $errors['warning_count'] === 0) {
+            if ($errors == false) {
                 $timestamp = $date->getTimestamp();
                 return $timestamp > 0 ? $timestamp : 0;
             }
         }
 
-        return 0;
+        return false;
```
ChrisCarini, line 213: should this change too? if so, it might be good to update the doc string

AH-DL: I think so as it is a static value, docstring updated :+1:  .

ChrisCarini: I'm not that familiar with this code base, but my inclination is to think that int return type is correct, and not boolean - the other exit location appears to return an integer unless I am misunderstanding the method :)

AH-DL: Thanks for the atention! its enough with the conditional fix, opened new [PR](https://github.com/kanboard/kanboard/pull/5116) to address that.

Issue #4253, opened 2019-07-29
Duplicating / Cloning project loses external links on all tasks


jpretori:
### Actual behaviour

I have set up a project with lots of tasks, many of which have lists of subtasks and external links to resources on the internet.

When I duplicate this project (all the checkboxes are ticked during duplication process), it duplicates the tasks, the subtasks... but not the external links.

### Expected behaviour

When duplicating a project with everything turned on, EVERYTHING should be duplicated into the new project.  Specifically, the external links on tasks should go across too.

### Steps to reproduce

1. Create a new project
2. Create a task in the project
3. Create an external link in the new task
4. Go to project configuration
5. Click "Duplicate" in the left pane
6. Turn on all the checkboxes
7. Open the task in the cloned project, notice that it does not have any external links

### Screenshots
![Screenshot from 2019-07-29 21-14-05](https://user-images.githubusercontent.com/16192343/62075772-482eeb00-b246-11e9-9623-b711aca1eef5.png)
![Screenshot from 2019-07-29 21-15-36](https://user-images.githubusercontent.com/16192343/62075773-48c78180-b246-11e9-89cf-b3cc5fc2f4ef.png)
![Screenshot from 2019-07-29 21-15-58](https://user-images.githubusercontent.com/16192343/62075775-48c78180-b246-11e9-9147-1e41bd78c23e.png)
![Screenshot from 2019-07-29 21-16-11](https://user-images.githubusercontent.com/16192343/62075777-49601800-b246-11e9-9722-2ab1b396c2a3.png)

### Configuration

    Application version: 1.2.9
    PHP version: 7.2.20-1+0~20190710.23+debian8~1.gbp2428c5
    PHP SAPI: fpm-fcgi
    OS version: Linux 4.9.0-0.bpo.9-amd64
    Database driver: mysql
    Database version: 10.1.40-MariaDB-1~jessie
    Browser: Mozilla/5.0 (X11; Linux x86_64; rv:68.0) Gecko/20100101 Firefox/68.0

jpretori:
After a little more testing, I have realised that this is affecting more than just external links.

Attachments to tasks are also not going over during project duplication.

19007361:
I have experienced this as well, and would like a solution.

19007361:
What it also does, is change the order of some of the tasks in the board (not to different columns though)

19007361:
Seems like v1.2.11 fixes this

19007361:
[...]

PR #5173, opened 2023-03-16
feat: duplicate attachments & external links during task duplication & importing

mundry:
Do you follow the guidelines?

- [x] There is no regression

Hi,

this PR adds functionality to duplicate attached files and external links for a task being duplicated to another project.

This means, when importing tasks from another project, or duplicating a task into another project, the task's attached files and external links are duplicated along with other attributes, like title and description.

No files are duplicated on disk. Instead, a new entry in the `task_has_files` table is created for the duplicate task, pointing to the same file on disk. In the duplicate task the attachment retains its original uploader as creator.

When removing a task, or removing an attachment from a task, the underlying file is only removed from disk, if there are no other tasks using the same file as attachment.

This functionality does not apply when duplicating a task within the same project.

Furthermore, internal links to other tasks are not duplicated, either.

This PR addresses requested functionality described in #4253.

Let me know any feedback you might have on this.

Cheers!

### DIFF
--- a/app/Model/FileModel.php
+++ b/app/Model/FileModel.php
@@ -199,10 +199,15 @@ abstract class FileModel extends Base
             $this->fireDestructionEvent($file_id);
 
             $file = $this->getById($file_id);
-            $this->objectStorage->remove($file['path']);
 
-            if ($file['is_image'] == 1) {
-                $this->objectStorage->remove($this->getThumbnailPath($file['path']));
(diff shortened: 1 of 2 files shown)


mundry:
I'm currently looking into fixing the breaking tests.

Any pointers on how to make the ExternalLinkManager available would be greatly appreciated.

PR #5278, opened 2023-06-02
fix(test): adjust plugin dir test to work on released versions

jnahmias:
Do you follow the guidelines?

- [X] There is no regression


jnahmias:
> What problem does it solve?

This allows the plugin unit test to succeed when run from a released version of kanboard (instead of just the git repo HEAD).
This is used in the Debian CI/autopkgtest system to run all the tests when any dependent component is updated to make sure that there are no regressions.
You can see more about the kanboard debci runs at: https://ci.debian.net/packages/k/kanboard/

PR #5317, opened 2023-07-18
Add support for phpunit 10

athos-ribeiro:
Add support for phpunit 10 by applying the following changes due to deprecated/removed features:

* Methods such as getName and setMethod were removed/replaced. Use the upstream suggested substitutes for those (i.e., name, onlyMethod, addMethod).
* Test file names must now match the within declared class names.
* withConsecutive was removed with no direct substitute use callbacks to test the desired behaviors.

Do you follow the guidelines?

- [ ] There is no regression
- [ ] I follow the existing [coding style](https://docs.kanboard.org/v1/dev/coding_standards/)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
     "require-dev": {
         "doctrine/instantiator": "1.5.0",
         "symfony/stopwatch": "5.4.21",
-        "phpunit/phpunit": "9.6.10",
+        "phpunit/phpunit": "^10",
         "roave/security-advisories": "dev-master"
     }
 }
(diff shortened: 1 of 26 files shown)


athos-ribeiro:
These seem to be the required changes to move into phpunit 10. However, some of them are breaking/not phpunit 9 compatible.

athos-ribeiro:
I am marking this as a draft due to the breaking changes it introduces. While I do not expect this to be merged ATM (due to the breaking changes), I suppose this should be incorporated at some point. Thoughts?

composer.json
```diff
@@ -64,7 +64,7 @@
     "require-dev": {
         "doctrine/instantiator": "1.5.0",
         "symfony/stopwatch": "5.4.21",
-        "phpunit/phpunit": "9.6.10",
+        "phpunit/phpunit": "^10",
```
xoxefdp, line 67: I think for production purposes it is required to set specific lib version (not caret "^")

PR #5330, opened 2023-08-16
favicon: make background transparent

atalii:
This is just a tiny change to make the icon look a bit better when on a desktop or a dark-mode browser. I just used GIMP's color to alpha tool for the ico and png - maybe there's a better way to do this, but it seems to work well enough.

By the way, I'm going to add a web manifest so I can put my instance on my phone's homescreen. Would a PR to add that upstream be appreciated as well? Thanks!

---

Do you follow the guidelines?

- [x] There is no regression
- [x] I have updated the unit tests and integration tests accordingly (not applicable)
- [x] I follow the existing [coding style](https://docs.kanboard.org/v1/dev/coding_standards/) (not applicable)


atalii:
> Somehow, I have the feeling that the new icon is less legible (not enough contrast). After looking on the web for best practices, I found this article https://web.dev/building-an-adaptive-favicon/
> 
> After experimenting, I was able to submit PR #5331 and that's seems to provide a better result:
> 

Yeah, I had the same concern, but felt that the icon was small enough to ignore. I had no idea that adaptive favicons were possible, though; your PR looks much better!

> If you don't mind, I would suggest to update the `favicon.png|ico` files with the same color scheme. You could use the file `inkscape-optimized-icon.svg` as source.

Sorry, but could you clarify this? I'm not quite sure what color scheme you're referring to or what hasn't been fixed by #5331. Thanks so much!

Issue #4963, opened 2022-02-22
2FA login is not center but left align


vins345:
### Checklist

- [x] I verified that Kanboard is correctly installed
- [x] I verified that the problem does not come from a plugin
- [x] I verified that the problem is not already reported
- [x] I understand that Kanboard is in maintenance mode. It doesn't mean it's abandoned, but there is no significant feature development

### Actual behaviour

2FA page login is align to left.

### Expected behaviour

2FA page align to center like normal login and pass form.

### Steps to reproduce
1. Turn on 2FA on your account
2. Try login

### Screenshots

![image](https://user-images.githubusercontent.com/91678388/155119893-045515c5-315e-456e-910d-f5615e4a813c.png)

### Configuration

Application version: v1.2.22
PHP version: 8.0.14
PHP SAPI: fpm-fcgi
HTTP Client: cURL
OS version: Linux 5.10.0-11-amd64
Database driver: sqlite
Database version: 3.36.0
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.102 Safari/537.36

auge8472:
I encountered the same problem during development of my own theme. I was able to reformat the login form but in the next step I found no point to apply my own CSS-rules to. IMHO it would be enough to provide an ID for the 2FA-form or (IMHO better) to apply the class `.login-form` also to the 2FA-form.

PR #5350, opened 2023-09-25
Put the 2FA-form into the center of the browser viewport

auge8472:
During login it is a bit irritating, to see the login form in the center of the browser viewport and to have to search for the two factor authentication form in the left upper corner of the viewport afterwards. Because of that I searched for a way to center the 2FA-form in the same way as the login form and found it by simply replacing the class for the complete container of the form with the same value as the one for the login form. This makes the two factor authentication form appearing in the middle of the screen like the login form does.

I left the similar form in `app/Template/twofactor/show.php` unchanged because AFAIK this is the form in the user settings, that is part of a more complex structure. (IMHO) That structure doesn't need to be changed.

This solves #4963.

Do you follow the guidelines?

- [x] There is no regression
- [ ] ~~I have updated the unit tests and integration tests accordingly~~ (AFAIK there is no need)

### DIFF
--- a/app/Template/twofactor/check.php
+++ b/app/Template/twofactor/check.php
@@ -1,4 +1,4 @@
-<section class="page">
+<section class="form-login">
     <div class="page-header">
         <h2><?= t('Two factor authentication') ?></h2>
     </div>


auge8472:
Thank you. :-)

PR #5420, opened 2024-02-20
Expand search bar. Fixes #4990

### LINKED ISSUE #4990: Search bar becomes too narrow when adding plugins (v.1.22 in english only)
### Checklist

### Actual behaviour

If you have several plugins installed that give extra buttons in the search/filter area, the search bar gets too narrow and becomes impossible to use (you have to copy/paste things in and out.)

This happens in english after updating from v1.2.20 to v.1.2.22

2022_04_06_15:56:36_Selection_001

After removing the plugin "Due Date" that creates this wide "Sorted by board order" button, the search bar is wide and can be used.

2022_04_06_15:51:03_Selection_001

### Expected behaviour

What we had in v.1.20 and still in v.1.2.22 if we switch the language in French, is that when there are too many buttons the search area will go in a separate line, with plenty of space.

The capture is from v.1.2.22 in French

2022_04_06_16:05:38_Selection_001

### Steps to reproduce

- Choose English as application language
- Install several plugins that add buttons (Calendar, Gantt, Due Date)

### Screenshots

Screenshots included above.

### Logs

No logs

### Configuration

    Application version: 1.2.22
    PHP version: 8.1.3
    PHP SAPI: apache2handler
    HTTP Client: cURL
    OS version: Linux 5.4.0-105-generic
    Database driver: mysql
    Database version: 10.3.13-MariaDB
    Browser: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:98.0) Gecko/20100101 Firefox/98.0

phkam:
Expand search bar when some plugins or other labels take up some space. Fixes #4990 

Do you follow the guidelines?

- [x] There is no regression


smirta:
Sorry for the spreading around @phkam . 

> @smirta, thanks for your message.
> The `margin-bottom` was to add more space underneath the search bar. But it
> might be possible that this is no longer required. What's the version of kanboard
> you noticed this issue with?
> 
> Kind Regards,
> Phil
> 

I'm somehow unable to reply to your comment in the fix. We're currently using 1.2.51. If this isn't required anymore, I'd prepare a PR which removes the margin. It currently leads to the misaligned display as shown in the screenshot with wider screens / higher resolution if not "enough" plugins with long names are installed :)

phkam:
@smirta 



> Sorry for the spreading around @phkam .

no problem; replying here is much easier.

 
> > The `margin-bottom` was to add more space underneath the search bar. But it
> > might be possible that this is no longer required. What's the version of kanboard
> > you noticed this issue with?
> 
> We're currently using 1.2.51. 
> If this isn't required anymore, I'd prepare a PR which removes the margin.
> It currently leads to the misaligned display as shown in the screenshot with
> wider screens / higher resolution if not "enough" plugins with long names are installed :)

I did some brief tests with Kanboard 1.2.51 and I ran into the display issue depicted in your attached screenshot.
However the issue appeared for me only when I disabled the margin. I had no issues out of the box with the default settings.

I am a bit puzzled how and why this is happening for you. 

I think it would be best to open a new issue/bugreport. Feel free to mention me there and include all the plugins you use and steps to reproduce the display issue.

smirta:
@phkam I forgot to mention that the display is different from browser to browser. I could now narrow it down to an issue with the font settings in the browser. E.g. the zen browser (a firefox fork) is displaying it misaligned but the exact same plain firefox isn't. So I started to have a look at the settings and discovered, that under Fonts in General the selected fonts are different. If I change them to the same as in the plain firefox (Noto Serif, Noto Sans and Noto Sans Mono), the page gets displayed correctly. The same applies also for Chrome/Chromium. I have a feeling that we'll have similar results with Edge and Safari.

I'll open an issue but I'm not enough web developer to tackle that issue.

smirta:
@phkam I'm so so sorry for having wasted your time. My profound apologies. It is quite embarrassing. I think I found the issue and it seems it was entirely my/our mistake. There was a custom CSS in place which is causing that.

phkam:
@smirta No worries! I’m glad you found the issue.

PR #5440, opened 2024-03-24
Made the max projects number configurable via config.php

Tagirijus:
I wanted to extend the number of projects shown on the dashboard, yet was not able to write a plugin for it, since it would require to overwrite the _DashboardController_, which - AFAIK - is not possible. So I thought I add global variable to make use of the `max` argument of the `getDashboardPaginator()` method and make a PR, in case you might find it usefull. It's just a very tiny change, I would say. Thanks for reading!

Do you follow the guidelines?

- [x] There is no regression


Tagirijus:
Thanks for pointing out. I was not aware of these environment variables. Good to know and great to learn new things. I hope my last commit did the trick then. Also: sorry for my late reply!

PR #5444, opened 2024-04-01
Added possibility to set task expenses to manage project expenses

Kaerol:
Do you follow the guidelines?

- [v ] I have tested my changes
- [v ] There is no breaking change
- [v ] There is no regression
- [x ] I have updated the unit tests and integration tests accordingly
- [ v] I follow the existing [coding style](https://docs.kanboard.org/v1/dev/coding_standards/)


Kaerol:
Hi!
I installed plugin you mention but it doesn't meet my requirments.

Agree that this feature is specyfic and maybe not fits to all projects - but from other hand there are other features which are used only in some project 😉 and... this feature may be the one which is silent required 😁

Rewrite it into plugin is a quite good sugestion - I thought about it but I couldn't find a solution how to add some HTML tags into existing templates - are ther any hooks? I saw some events but hooks 🤔

Another propsal is add some checkbox to hide this feature if not needed 😉

Kaerol:
Ok, thanks.
I will take a look and check it can be helpful for me.

PR #5449, opened 2024-04-09
Add role privacy to comments

JustRaosha:
This is a new feature that enables privileged users to exchange messages hidden to less privileged ones.

This is my first time contributing to this project, so feel free to make constructive criticism. I did not write any test yet, so if you feel it is necessary, please advise on that as well.

Greetings.

- [X] There is no regression

### DIFF
--- a/app/Api/Authorization/CommentAuthorization.php
+++ b/app/Api/Authorization/CommentAuthorization.php
@@ -2,6 +2,9 @@
 
 namespace Kanboard\Api\Authorization;
 
+use JsonRPC\Exception\AccessDeniedException;
+use Kanboard\Core\Security\Role;
+
 /**
  * Class CommentAuthorization
  *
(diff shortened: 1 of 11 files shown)


JustRaosha:
I am using application roles because custom roles need new projects restrictions to be added (‘Project manager comments view is not permitted’, …). We also saw a use case where clients would only have app-user role and devs app-manager role so it could be enough to use application roles to restrict the comment privacy and allow the devs to share information between them about the task, avoiding disturbing the client with unnecessary ones.

JustRaosha:
Would you mind answering my previous comment so that I know if I go develop again or not ?

JustRaosha:
Thank you really much for all of these feedbacks, I'm currently working on this.

lazka:
It looks like this broke sending comments via email: https://github.com/kanboard/kanboard/issues/5486

PR #5455, opened 2024-04-18
Fix various action executions failing if the duration parameter isn't set

### LINKED ISSUE #5454: Cron fails if an action is created without setting a duration
### Checklist

### Actual behaviour

cron fails with:

```php
PHP Fatal error:  Uncaught TypeError: Unsupported operand types: string * int in /application/app/Action/TaskEmailNoActivity.php:87
Stack trace:
#0 /application/app/Action/Base.php(269): Kanboard\Action\TaskEmailNoActivity->doAction()
#1 /application/vendor/symfony/event-dispatcher/EventDispatcher.php(230): Kanboard\Action\Base->execute()
#2 /application/vendor/symfony/event-dispatcher/EventDispatcher.php(59): Symfony\Component\EventDispatcher\EventDispatcher->callListeners()
#3 /application/app/Console/TaskTriggerCommand.php(50): Symfony\Component\EventDispatcher\EventDispatcher->dispatch()
#4 /application/app/Console/TaskTriggerCommand.php(27): Kanboard\Console\TaskTriggerCommand->sendEvent()
#5 /application/vendor/symfony/console/Command/Command.php(298): Kanboard\Console\TaskTriggerCommand->execute()
#6 /application/app/Console/CronjobCommand.php(29): Symfony\Component\Console\Command\Command->run()
#7 /application/vendor/symfony/console/Command/Command.php(298): Kanboard\Console\CronjobCommand->execute()
#8 /application/vendor/symfony/console/Application.php(1040): Symfony\Component\Console\Command\Command->run()
#9 /application/vendor/symfony/console/Application.php(301): Symfony\Component\Console\Application->doRunCommand()
#10 /application/vendor/symfony/console/Application.php(171): Symfony\Component\Console\Application->doRun()
#11 /application/cli(21): Symfony\Component\Console\Application->run()
#12 {main}
  thrown in /application/app/Action/TaskEmailNoActivity.php on line 87
```

### Expected behaviour

Running cron should succeed.

### Steps to reproduce

The problem is caused when a user creates an action that has an empty duration. This gets then stored in the the `actions_has_params` as an empty string. When the action is executed it gets multiplied by an integer https://github.com/kanboard/kanboard/blob/8505589566714b7e47175703d4c2c28eece4d776/app/Action/TaskEmailNoActivity.php#L87 which fails with the error above.

Steps:

* Go to project
* Go to "Configure this project"
* Go to "Automatic actions"
* Click "Add a new action"
* Select "Send email when there is no activity on a task"
* Click "Next Step"
* Click "Next Step"
* Now three inputs appear, leave them empty and click "Save"
* On the server run `./cli cronjob` and see it fail

This affects the following actions at least (all that have a duration parameter):

* 'Send email when there is no activity on a task'
* 'Close a task when there is no activity in a specific column'
* 'Close a task in a specific column when not moved during a given period'
* 'Send email when there is no activity on a task'
* 'Move the task to another column when not moved during a given period'
* 'Move the task to another column when the due date is less than a certain number of days'

### Configuration

- Kanboard version: 1.2.35
- Database type and version: sqlite
- PHP version: 8.2.18
- OS: Debian Bookwork
- Browser: Firefox

lazka:
It's easy to create an action that has nothing set for the duration by just going "next" in the UI. This results in an empty string being stored in the DB for the duration, which with PHP 8 fails as it is multiplied by an integer during the action execution, resulting in:

TypeError: Unsupported operand types: string * int

With older PHP this resulted in a warning and "0" as a result. By casting to int explicitely we restore that behaviour and we get 0 if the string can't be converted.

Fixes #5454

Do you follow the guidelines?

- [x] There is no regression

### DIFF
--- a/app/Action/TaskCloseNoActivity.php
+++ b/app/Action/TaskCloseNoActivity.php
@@ -68,7 +68,7 @@ class TaskCloseNoActivity extends Base
     public function doAction(array $data)
     {
         $results = array();
-        $max = $this->getParam('duration') * 86400;
+        $max = (int)$this->getParam('duration') * 86400;
 
         foreach ($data['tasks'] as $task) {
             $duration = time() - $task['date_modification'];
--- a/app/Action/TaskCloseNoActivityColumn.php
+++ b/app/Action/TaskCloseNoActivityColumn.php
@@ -69,7 +69,7 @@ class TaskCloseNoActivityColumn extends Base
     public function doAction(array $data)
     {
         $results = array();
-        $max = $this->getParam('duration') * 86400;
+        $max = (int)$this->getParam('duration') * 86400;
 
         foreach ($data['tasks'] as $task) {
             $duration = time() - $task['date_modification'];
--- a/app/Action/TaskCloseNotMovedColumn.php
+++ b/app/Action/TaskCloseNotMovedColumn.php
@@ -69,7 +69,7 @@ class TaskCloseNotMovedColumn extends Base
     public function doAction(array $data)
     {
         $results = array();
-        $max = $this->getParam('duration') * 86400;
+        $max = (int)$this->getParam('duration') * 86400;
 
         foreach ($data['tasks'] as $task) {
             $duration = time() - $task['date_moved'];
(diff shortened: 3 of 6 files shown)


lazka:
Thanks!

PR #5465, opened 2024-05-14
Add link to "Add a comment" in the comments section before the "Change sorting"

temp5mr:
1. It helps when the comment input field is far away (after several comments).
2. It's intuitively easier than searching for an item in the left panel.
3. It's convenient specifically in combination with "Sorting": you look at the first/last and then write.


app/Template/task_comments/show.php
temp5mr, line 1: commentaddlink

PR #5474, opened 2024-05-31
Update German translations.php

onlygecko:
Added translation for 'Close all tasks of this column'

Do you follow the guidelines?

- [x] There is no regression


onlygecko:
Thanks for the quick reply. :smile:

And thanks for the hint. :pray: The Greenwing plugin is responsible for the missing translation. Since the plugin is not actively maintained, we will deal with this in other way.

PR #5489, opened 2024-06-12
Added comment visibility check in activity

YannisHoareau:
This should fix shown comments in activity that roles were not allowed to see in other contexts.

Do you follow the guidelines?

- [x] There is no regression


YannisHoareau:
Changed the way it worked, I hope that was what you thought. Tests done on Kanboard web and with API.

PR #5503, opened 2024-06-22
A rather big update for the l10n for the Greek language

lourdas:
Do you follow the guidelines?

- [X] There is no regression

I've been considering Kanboard for our use case and the Greek language had typos, missing words, wrong meaning in a few translated phrases, etc. and there were also a lot of new untranslated messages. I hopefully took care of them all.

Please consider merging in mainline.

Thank you!


lourdas:
The tests run and came up with a missing %s in a string, which I included in the second commit ([115105c](https://github.com/kanboard/kanboard/pull/5503/commits/115105cc948957fc829f5a591af8c07ed2b9fbe7)).

PR #5559, opened 2024-09-20
Move subtask to another task

arocholl:
Do you follow the guidelines?

- [ x] I have tested my changes
- [ x] There is no breaking change
- [x ] There is no regression
- [x ] I follow the existing [coding style](https://docs.kanboard.org/v1/dev/coding_standards/)
- [ x] Ideally, my commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/)

Changes were tested manually to confirm only the task ID is changed accordingly

### DIFF
--- a/app/Controller/SubtaskController.php
+++ b/app/Controller/SubtaskController.php
@@ -219,4 +219,52 @@ class SubtaskController extends BaseController
             throw new AccessForbiddenException();
         }
     }
+    
+    /**
+     * Handle the GET request to display the form to move a subtask to another task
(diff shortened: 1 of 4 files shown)


arocholl:
> Your pull request does not account for project permissions. It currently allows anyone to move any subtask to any task, regardless of the configured access controls. This may be considered a security issue by some people.

I see, I will check if I can add this into the code change. I based my changes in subtask edit and I think it was equivalent in all regards.

PR #5630, opened 2025-02-28
Fix #5401 internally linked tasks' title no longer wrap under dropmenu icon.

### LINKED ISSUE #5401: Internally linked task cards with long titles have incorrectly wrapping text
### Checklist

### Actual behaviour

When a row item is too long in the "Internal links" section of a task's details, it will wrap the text completely to the left of the row, without taking the "⚙️⏷" into account.

In comparison, this does not affect the "Subtasks" table, which correctly wraps the text after the "⤭⚙️⏷☐" section.

### Expected behaviour

Line wrapping should be aligned to be after the "⚙️⏷"; in other words, it should behave like the nice "Subtasks" table's behavior.

I haven't tested the "External links" feature, but presumably it might be affected by the same issue too?

### Steps to reproduce

Create a task card that has internal links to other task cards that have long titles, and watch the line wrap below its gear icon.

### Screenshots

Circled in red is the incorrect behavior, circled in green is the expected behavior:

image

### Configuration

- Kanboard version: 1.2.33 and 1.2.34
- Database type and version: SQLite 3.40.1
- PHP version: 8.2.14
- OS: Fedora 39
- Browser: Firefox 121

Al-27:
#5401 
- [x] There is no regression
<del> - [ ] I have updated the unit tests and integration tests accordingly </del>
- [x] Ideally, my commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/)


app/Template/task_internal_link/table.php
```diff
@@ -17,41 +17,43 @@
...
+                                        </li>
+                                        <li>
+                                            <?= $this->modal->confirm('trash-o', t('Remove'), 'TaskInternalLinkController', 'confirm', array('link_id' => $link['id'], 'task_id' => $task['id'])) ?>
+                                        </li>
+                                    </ul>
+                                </div>
+                            <?php endif ?>                         
+                        <?php endif ?>
+
+                    <div>
```
Al-27, line 44: reason for wrapping `url->link` and `text->e` in the same element, is to avoid having three flex items, which could've resulted in something like this
image

that's why i added the `$link_mode` and `$link_array`, and put the `url->link` function outside the condition body.

assets/css/src/dropdown.css
```diff
@@ -74,6 +74,10 @@ ul.dropdown-submenu-open {
     text-decoration: none
 }
 
+.dropdown-menu-link-icon {
+    display: inline-flex
```
Al-27, line 78: used `inline-flex`  instead of `flex`, as it would break the external links' title
image

assets/css/src/task_links.css
```diff
@@ -9,4 +9,8 @@
 
 .task-link-closed {
     text-decoration: line-through
-}
\ No newline at end of file
+}
+
+.task-links-table-td {
```
Al-27, line 14: following in the footsteps of [identei](#4366)

PR #5632, opened 2025-03-03
feat: _add_ tags to tasks, instead of replacing all tags in bulk. Fixes #4952

### LINKED ISSUE #4952: Bulk _add_ tags to tasks, instead of replacing all tags
- [x] I understand that Kanboard is in maintenance mode. It doesn't mean it's abandoned, but there is no significant feature development.
- [X] I have read this document: https://docs.kanboard.org/en/latest/developer_guide/faq.html#why-are-you-not-developing-my-feature-request

When editing tasks in bulk, I can _change_ **all** tags, but I cannot just _add_ tags. I would really appreciate this feature.

Screenshot 2022-02-11 at 09 44 02
Screenshot 2022-02-11 at 09 47 07

Heart1010:
Fixes #4952

image


ianme:
Great, thank you!

PR #5661, opened 2025-05-15
Added minor docker improvements

DaanSelen:

- [ x] I have tested my changes thoroughly
- [ x] No breaking changes were introduced
- [ x] No regressions were observed
- [ x] I have updated the unit tests and integration tests accordingly
- [ x] I follow the existing [coding standards](https://docs.kanboard.org/v1/dev/coding_standards/)
- [ x] My commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/)

I have added a healthcheck, which checks for a HTTP 200 code on the /login page. Which should indicate if something is very wrong when it returns something else.

Removed version indication and default port 80 exposure, leaving encrypted HTTP as the first default option. Does not regress just make the default more secure.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,21 +12,24 @@ VOLUME ["/var/www/app/data", "/var/www/app/plugins", "/etc/nginx/ssl"]
 
 EXPOSE 80 443
 
-ARG VERSION
+ARG VERSION="latest"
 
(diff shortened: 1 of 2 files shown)


DaanSelen:
Yes I see, thanks! I am just seeing the healthcheck.php! Hope this helps. I was wondering what happend!

PR #5717, opened 2025-09-08
feat!: support RTL

LaisRast:
This pull request introduces basic support for right-to-left (RTL) languages in Kanboard.
Users selecting Arabic or Farsi will now see the interface rendered in right-to-left mode, improving usability and readability. I have marked this PR as a breaking change since it alters the rendering for these languages.

This lays the groundwork for enhanced RTL support across the application and future contributions.
Non-RTL languages remain unaffected.

- [X] I have tested my changes thoroughly
- [ ] No breaking changes were introduced
- [X] No regressions were observed
- [X] I follow the existing [coding standards](https://docs.kanboard.org/v1/dev/coding_standards/)
- [X] My commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/)

### DIFF
--- a/app/Helper/AppHelper.php
+++ b/app/Helper/AppHelper.php
@@ -131,6 +131,17 @@ class AppHelper extends Base
     }
 
     /**
+     * Check if current language requires RTL direction
+     *
+     * @access public
+     * @return bool
+     */
+    public function isRtlLanguage()
+    {
+        return $this->languageModel->isRtlLanguage();
+    }
+
+    /**
      * Get date format for Jquery DatePicker
      *
      * @access public
--- a/app/Model/LanguageModel.php
+++ b/app/Model/LanguageModel.php
@@ -194,6 +194,24 @@ class LanguageModel extends Base
     }
 
     /**
+     * Check if current language requires RTL direction
+     *
+     * @access public
+     * @return bool
+     */
+    public function isRtlLanguage()
+    {
+        $rtlJsLanguageCodes = array(
+            'ar',
+            'fa',
+        );
+
+        $lang = $this->getJsLanguageCode();
+
+        return in_array($lang, $rtlJsLanguageCodes);
+    }
+
+    /**
      * Get current language
      *
      * @access public
--- a/app/Template/layout.php
+++ b/app/Template/layout.php
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="<?= $this->app->jsLang() ?>">
+<html lang="<?= $this->app->jsLang() ?>"<?php if ($this->app->isRtlLanguage()): ?> dir="rtl"<?php endif; ?>>
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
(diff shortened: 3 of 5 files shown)


LaisRast:
> Wouldn’t it be easier to use `dir="auto"` and let the browser handle that automatically?

That would certainly be simpler, but it’s not reliable. `dir="auto"` is useful for small snippets of unknown or mixed text, but not for the root `<html>`. For the whole document, [W3C recommends explicitly setting `dir="rtl"` if the overall direction is right-to-left](https://www.w3.org/International/questions/qa-html-dir). This ensures consistent layout and accessibility. Since we already know the language in the app, keeping the explicit `dir` is the more robust choice.

Why isn’t `dir="auto"` reliable on the `<html>` tag?  
The [documentation for `dir`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/dir) explains:

> `auto`, which lets the user agent decide. It uses a basic algorithm as it parses the characters inside the element until it finds a character with a strong directionality, then applies that directionality to the whole element.

The browser has to scan for the first strong character, which may not be representative (e.g., a text logo or non-translated application name). In Kanboard, the first strong character in the base theme is the title `header .title-container .title`. If that title isn’t translated, or if the translation begins with a non-RTL character for any reason, the entire layout direction could be set incorrectly.

auge8472:
Even I'm being late to the party, as someone who has redesigned the web interface of their own project to be writing direction independent, I would like to share a few experiences here. The current implementation supports only two rtl written languages (Arabic and Farsi). Supporting i.e. Hebrew would need further changes in the code.

For future enhancements I would recommend to introduce a new language key (something like `"writing direction"`) with the possible values "ltr" and "rtl" (are there languages with *mandatory* vertical writing directions on this planet?). Adding the key to all languages would make the check for `<?php if ($this->app->isRtlLanguage()): ?> dir="rtl"<?php endif; ?>` in `app/Template/layout.php` obsolete. Maybe a safety net is needed for the case of a language file with no key for writing-direction.

Furthermore there is much work left. Direction dependent CSS properties like `margin-left` or `padding-right` will often not do their intended job in rtl languages. There are logical properties which can replace the direction dependent properties.

As an example take the sub properties of `padding` and keep ltr written languages in mind.

~~~
padding-top => padding-block-start
padding-right => padding-inline-end
padding-bottom => padding-block-end
padding-left => padding-inline-start

padding-inline (for start and end with one or two different values)
padding-block (for start and end with one or two different values)
~~~

Noting `padding-left` for an elements content (i.e. a link text) in ltr written languages means you'll add space between the content and the edge of the element *in front* of the content. In rtl written languages this space will be on the wrong side *behind* the content.

With the logical properties you'll have at least two advantages. Firstly using `padding-inline-start` will automatically put the space in front of the content in every writing direction. And second you can rotate an element without losing the link between the side of an element and the property. The same applies to `margin` and several further CSS properties.

Reading the following articles helped me a lot.

- [MDN: CSS logical properties and values](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values)
- [CSS-Tricks: CSS Logical Properties and Values](https://css-tricks.com/css-logical-properties-and-values/)

PR #5747, opened 2025-12-25
Fix: Resolve SQL cardinality violation when a task has multiple links

ArsenKurd:
In the extended task query used by the board view, the subquery for the is_milestone flag used SELECT DISTINCT 1 without limiting the result set. If a task was linked more than once to a milestone (via duplicate or multiple task_has_links entries with the label 'is a milestone of'), the subquery returned multiple rows, causing MySQL to raise a "Subquery returns more than 1 row" error (SQLSTATE 21000).

This change adds LIMIT 1 to the subquery, ensuring it always returns at most one row (1 if any such link exists, NULL otherwise), which satisfies the scalar context requirement while preserving the intended logic. The fix is compatible with both MySQL and PostgreSQL and has no performance impact—thanks to early termination upon finding the first matching record.

- [X] I have tested my changes thoroughly
- [X] No breaking changes were introduced
- [X] No regressions were observed
- [X] I follow the existing [coding standards](https://docs.kanboard.org/v1/dev/coding_standards/)
- [X] My commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/)

### DIFF
--- a/app/Model/TaskFinderModel.php
+++ b/app/Model/TaskFinderModel.php
@@ -86,7 +86,7 @@ class TaskFinderModel extends Base
                 '(SELECT COUNT(*) FROM '.SubtaskModel::TABLE.' WHERE '.SubtaskModel::TABLE.'.task_id=tasks.id AND status=2) AS nb_completed_subtasks',
                 '(SELECT COUNT(*) FROM '.TaskLinkModel::TABLE.' WHERE '.TaskLinkModel::TABLE.'.task_id = tasks.id) AS nb_links',
                 '(SELECT COUNT(*) FROM '.TaskExternalLinkModel::TABLE.' WHERE '.TaskExternalLinkModel::TABLE.'.task_id = tasks.id) AS nb_external_links',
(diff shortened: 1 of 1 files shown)


fguillot:
What version of Mysql are you using? I'm not able to reproduce locally with MariaDB.

Assuming a task with multiple `is a milestone of` relationships:

```sql
MariaDB [kanboard]> select * from task_has_links where task_id=1;
+----+---------+---------+------------------+
| id | link_id | task_id | opposite_task_id |
+----+---------+---------+------------------+
|  1 |       8 |       1 |                2 |
|  3 |       8 |       1 |                3 |
+----+---------+---------+------------------+
2 rows in set (0.004 sec)
```

Running the subquery with the `DISTINCT` returns only one row for me:

```sql
MariaDB [kanboard]> select distinct 1 from task_has_links tl join links l on tl.link_id = l.id where tl.task_id = 1 and l.label = 'is a milestone of';
+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.002 sec)
```

It returns 2 rows without the `DISTINCT` (as expected):

```sql
MariaDB [kanboard]> select 1 from task_has_links tl join links l on tl.link_id = l.id where tl.task_id = 1 and l.label = 'is a milestone of';
+---+
| 1 |
+---+
| 1 |
| 1 |
+---+
2 rows in set (0.001 sec)
```

ArsenKurd:
> What version of Mysql are you using?
I'm running MySQL 8.0.15 for a long time.

PR #5774, opened 2026-02-13
fix(css): Fix accessibility issue of insufficient text to background contrast

umweltinfo:
While checking for WCAG guideline 1.4.3 (AA, Must) a few UI elements showed insufficient contrast.

~~This change tries to resolve this without fundamental theming changes by replacing color-light with color-medium in a few places.~~

This change tries to resolve this by making `--color-light` slightly darker but still lighter than `--color-medium` in the light theme.

This will bring contrast up to 4.48 which is almost 4.5 as required by the WCAG guideline while still trying at being minimal invasive.


adamreichold:
> This change makes dark mode harder to read.

Would it be alright if we tried to change the definition `--color-light` itself then to achieve sufficient contrast in the light theme?

PR #5804, opened 2026-04-18
Escape notification title in web notification template

TristanInSec:
## Summary

Escape the notification title using `TextHelper::e()` in the web notification
template to prevent HTML injection via crafted task titles or filenames.

Only the notification template is modified. `UrlHelper::link()` is left
unchanged.

### DIFF
--- a/app/Template/web_notification/show.php
+++ b/app/Template/web_notification/show.php
@@ -53,9 +53,9 @@
             <?php endif ?>
 
             <?php if ($this->text->contains($notification['event_name'], 'task.overdue') && count($notification['event_data']['tasks']) > 1): ?>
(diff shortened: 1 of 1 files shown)


fguillot (changes requested):
- This change completely break the UI and the text is now escaped twice
- There is already a method for escaping: `TextHelper:e()`
- There is no unit tests for the new method `linkRaw`

Image

TristanInSec:
Hi @fguillot, thank you for the review. You are right, modifying link() was the wrong approach. I reworked the PR to escape only the notification title at the template level using TextHelper::e(), which is the actual injection point. UrlHelper::link() is no longer touched.

PR #5806, opened 2026-04-18
Validate user_id in subtask time tracking write methods

TristanInSec:
## Summary

Add user_id ownership check to the write methods `setSubtaskStartTime()`
and `setSubtaskEndTime()` to prevent project members from starting or
stopping timers attributed to other users.

The read methods `hasSubtaskTimer()` and `getSubtaskTimeSpent()` are left
unchanged as they are legitimate cross-user reads for project members.


fguillot (changes requested):
`hasSubtaskTimer()` and `getSubtaskTimeSpent()` are legitimate cross-user reads for anyone with project access.

TristanInSec:
Hi @fguillot, thank you for the clarification. I removed the check from hasSubtaskTimer() and getSubtaskTimeSpent() since cross-user reads make sense for project visibility.

The updated PR only adds the user_id validation to the two write methods (setSubtaskStartTime and setSubtaskEndTime) to prevent a project member from starting or stopping timers attributed to another user.

PR #5816, opened 2026-05-11
fix: Adjusted tooltip max height and position

jamie-belanger:

Made two cosmetic changes to the tooltip popup to address two annoyances I encounter frequently.

First, limited the max height to 50% of the viewport (I apparently write too much in my descriptions).

Second, I adjusted the positioning on screen so tooltips no longer open downward off the bottom of the screen.

The problem:
problem

And after this change:
problem-fixed


assets/js/core/tooltip.js
```diff
@@ -61,10 +61,11 @@ KB.on('dom.ready', function() {
         containerElement.addEventListener("mouseenter", mouseOnTooltip, false);
         containerElement.mouseOnTooltip = false;
 
-        var elementRect = element.getBoundingClientRect();
-        var top = elementRect.top + window.scrollY + elementRect.height;
-        containerElement.style.top = top - 20 + "px";
+        // Elements don't have height until rendered by DOM, so add as hidden so it doesn't flicker
+        containerElement.style.visibility = false;
```
fguillot: `style.visibility` only accepts the strings "hidden", "visible", or "collapse". Assigning a boolean is coerced to "false"/"true", which are invalid CSS values and the browser silently ignores them.

See https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/visibility

jamie-belanger: Good catch. Fixed.

PR #5818, opened 2026-05-17
feat: Log password authentication failures

Stephen-Seo:
Log password authentication failures with the selected `LOG_DRIVER` for integration with `fail2ban`.

An example regex for a `fail2ban` filter is as follows:
`failregex = ^.*Kanboard: Failed login from: <HOST> username:.*$`

This is my first time working with a `php` project and I expect that this PR may not be merged without more changes.

The only changes made were two additional `logger` log lines that I have tested on my RaspberryPI4 running `Kanboard`, so I am certain that even though ~~this PR wasn't tested more rigorously~~, I am of the opinion that this code should work as expected.

EDIT 2026-05-22: I've tested this by banning my phone and Kanboard does output login failures (via password) properly with these changes and fail2ban did pick up on the logs properly.

I felt this commit was necessary because `fail2ban` depends on log output to work, and I could not find existing documentation or options for integration with `fail2ban`.

I am open to discussion and changes if needed.

### DIFF
--- a/app/Validator/AuthValidator.php
+++ b/app/Validator/AuthValidator.php
@@ -62,6 +62,7 @@ class AuthValidator extends BaseValidator
         if ($this->userLockingModel->isLocked($values['username'])) {
             $result = false;
             $errors['login'] = t('Your account is locked for %d minutes', BRUTEFORCE_LOCKDOWN_DURATION);
+            $this->logger->error('Kanboard: Failed login from: ' . $_SERVER['REMOTE_ADDR'] . ' username: ' . $values['username']);
             $this->logger->error('Account locked: '.$values['username']);
         }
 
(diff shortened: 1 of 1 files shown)


fguillot (changes requested):
Kanboard already logs password authentication failures for fail2ban in `AuthSubscriber::onLoginFailure()` using `error_log()`.

So this change duplicates the failed-password event instead of moving that logging into the configured Kanboard logger.

If we want `LOG_DRIVER` support, the better place is probably `AuthSubscriber::onLoginFailure()`, where the auth-failure event is already handled.

Two other notes:

- `$_SERVER['REMOTE_ADDR']` skips `$this->request->getIpAddress()`, which respects `TRUSTED_PROXY_HEADERS` and trusted proxy networks. Otherwise fail2ban can ban the proxy/container IP.
- The `validateLocking()` line fires when an account is already locked, before password authentication runs, so it is not logging an actual failed password check.

Stephen-Seo:
After some checking, I realize I hadn't configured my server correctly. The `error_log = syslog` option wasn't set, and after setting this, the existing logs started working. Closing this pull request as the existing settings work. Sorry for the noise.

Stephen-Seo:
For future reference, a working filter line in fail2ban works with the existing implementation (when syslog is redirected to systemd-journal):

    failregex = ^.*kanboard\[[0-9]+\]:\s+Kanboard: user.*authentication failure with IP address: <HOST>$

PR #5830, opened 2026-06-04
Fix default notifications not being enabled for new users when system setting is on

RAOEUS:
On the System Settings administration page (`/?controller=ConfigController&action=application`), when the `Enable notifications by default for all new users` box is checked, new users' notifications are still disabled upon user creation and the user must visit their notification settings and click the `Save` button to enable notifications. This change ensures it follows that setting and enables the notifications upon user creation.

- [x] My commit messages follow the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/)

### DIFF
--- a/app/Controller/UserCreationController.php
+++ b/app/Controller/UserCreationController.php
@@ -69,6 +69,7 @@ class UserCreationController extends BaseController
             }
 
             if ($this->configModel->get('notifications_enabled', 0) == 1) {
+                $this->userNotificationModel->enableNotification($user_id);
                 $this->userNotificationTypeModel->saveSelectedTypes($user_id, [MailNotification::TYPE, WebNotification::TYPE]);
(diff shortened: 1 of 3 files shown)


fguillot:
That seems to work fine on my side. The template has a hidden field to enable the notifications when the global option is activated.

https://github.com/kanboard/kanboard/blob/28c31dd8fef56f9143161e280d1570d34493ba59/app/Template/user_creation/show.php#L57-L61

RAOEUS:
Ah, I see! I believe it was because the OAuth2 plugin was creating the users for me, and nothing in there was respecting that rule. I will investigate and see if I can patch it directly in there.

RAOEUS:
I believe I have fixed this issue with a PR to the OAuth2 plugin: kanboard/plugin-oauth2#53

Thank you for your time!

